PR #145, opened 2014-09-16
Fail on pylint errors

opalmer:
This is the same change that was merged in pyfarm/pyfarm-master#187 except for the agent.


opalmer:
I'm going to abandon this one for now.  It's really far behind master and there are other changes which are more important than this for the time being.

PR #185, opened 2014-12-14
Agent System ID -> UUID

opalmer:
This change transitions the agent from building a system identifier based on hardware information to using a UUID.  This pull request is dependent upon:

* pyfarm/pyfarm-core#35
* pyfarm/pyfarm-master#311

This change could be updated to retain the old `systemid` field in the request however I choose to break this because we're sending along a different type than we were before.


guidow:
Running tasks does not work with this. There is still a bug in here, which should be addressed by #186.

pyfarm/agent/entrypoints/main.py
```diff
@@ -128,15 +128,23 @@ def __init__(self):
...
+        global_network.add_argument(
+            "--agent-id-cache", config="agent_id_cache",
+            default=None,
+            help="The location to cache the value for --agent-uuid.  By "
+                 "default the agent's UUID will be cached in the first "
+                 "available location on the configuration path.  Declaring "
+                 "this flag will override this behavior and attempt to load "
+                 "and save the agent's UUID from this location instead.  If "
+                 "this flag is provided along with --agent-uuid the value "
+                 "loaded from the cache will be overwritten by the value "
```
guidow: This is a bit ambiguous. Will specifying both options overwrite the UUID cache on disk?

pyfarm/agent/utility.py
```diff
@@ -275,3 +273,127 @@ def total_seconds(td):
...
+        """
+        assert isinstance(agent_uuid, UUID), agent_uuid
+        cls.log.debug("Attempting to save UUID %s to disk.", agent_uuid)
+
+        if path is not None:
+            return cls._save(agent_uuid, path=path)
+
+        directories = config.directories(validate=False, unversioned_only=True)
+        for directory in directories:
+            path = cls._save(agent_uuid, join(directory, "uuid.dat"))
```
opalmer: Yeah agreed....very bad place to have that stored by default.  This should be fixed in d85aa8d387a474cfb4b4d1dd41b952f12a693a52 which also improved the error handling.

pyfarm/agent/utility.py
```diff
@@ -275,3 +273,134 @@ def total_seconds(td):
...
+        to the first available path in the configuration.
+        """
+        assert isinstance(agent_uuid, UUID), agent_uuid
+        cls.log.debug("Attempting to save UUID %s to disk.", agent_uuid)
+
+        if path is not None:
+            return cls._save(agent_uuid, path=path)
+
+        directories = config.directories(validate=False, unversioned_only=True)
+        directories.reverse()
```
guidow: There is still a problem with this that did not occur to me yesterday:

There is a very high likelyhood that this will end up choosing the current user's home directory. The system-wide configuration directory will likely not be writable, and the user's $HOME is next in the list.
The problem with this is, in large organisations, the home directories may well be network mounts or roaming profiles. (They are here.) This is a problem because then, different agent instances running on different machines but under the same user will have the same UUID.

We are going to have to rethink our strategy here. Going through ```config.directories()``` and saving the uuid in the first path where that works just isn't a good idea. There is no directory in there that is actually a good choice for that: The system-configuration directory is probably not writable, the user's home directory might be shared with other machines in some way and the current working directory is a bad choice in any case, because it could be anything.

My idea would be to have the full path to uuid.dat configured explicitly in the agent.yml config file (with some sensible defaults), and load the UUID and store it to this path, and this path only. Trying various different paths is just a potential source of confusion. (Example: An administrator wants to force a node to generate a new UUID for some reason. He finds one uuid.dat file and deletes it. For some reason, there is another uuid.dat in some other place that the agent still reads, so it retains its old uuid even though it isn't supposed to. If the administrator is lucky, they will at least notice that the planned UUID reset did not work.)
If we cannot access that path in the way we need to, just quit with an error message.

We might need different configuration keys for the differerent supported OS types (Linux, BSD, OS X, Windows), though.

Here's my idea for defaults:

- Linux, BSD: ```/etc/pyfarm-agent/uuid.dat``` (The user or the install script will have to make sure this directory exists and is writable by the pyfarm user.)
- Windows: ```%SYSTEMDRIVE%\pyfarm-agent\uuid.dat```
- OS X: No idea.

opalmer: I think you're right, putting this in the config is really the best place for it.  I completely agree on that but do think the paths need to be a little different:

* Linux, BSD: `/etc/pyfarm/agent/uuid.dat` (This will mirror where the config files normally live so we don't have /etc/pyfarm/agent and /etc/pyfarm-agent.  If you'd rather not do this, then we should probably update pyfarm.core so we're not nesting directories under /etc/pyfarm/)
* Windows: `%COMMONAPPDATA%\pyfarm\agent\uuid.dat` (The canonical location for common application data on Windows, another option could be under LOCALAPPDATA which shouldn't roam.  More info here: http://www.microsoft.com/security/portal/mmpc/shared/variables.aspx)
* OS X: `/Library/pyfarm/agent/uuid.dat` (The config object uses /Library/pyfarm already and most applications seem to use /Library already too)

guidow: > * Windows: `%COMMONAPPDATA%\pyfarm\agent\uuid.dat` (The canonical location for common application data on Windows, another option could be under LOCALAPPDATA which shouldn't roam.

On my Windows 7 install here, %COMMONAPPDATA% seems to be undefined, so I guess %LOCALAPPDATA% is it. This seems to be exactly what we need anyway.

pyfarm/agent/entrypoints/main.py
```diff
@@ -128,15 +128,23 @@ def __init__(self):
...
-            "--systemid-cache", config="agent_systemid_cache",
-            help="The location to cache the value for --systemid. "
+            "--agent-id", config="agent_id", type=uuid_type,
+            default=AgentUUID.load(),
+            help="The UUID used to identify this agent to the master.  By "
+                 "default the agent will attempt to load a cached value "
+                 "however a specific UUID could be provided with this flag. "
                  "[default: %(default)s]")
+        global_network.add_argument(
+            "--agent-id-cache", config="agent_id_cache",
```
guidow: I think we should rename this to agent-id-store or agent-id-file. A cache is a performance optimisation. It stores data closer to where it will likely be needed, and it usually can be deleted/cleared/flushed/whatever without breaking anything.

The uuid.dat file OTOH has nothing to do with performance, and deleting it will break things.

pyfarm/agent/utility.py
```diff
@@ -275,3 +273,134 @@ def total_seconds(td):
...
+
+            return UUID(data)
+
+        except (OSError, IOError) as e:
+            if e.errno != ENOENT:
+                raise
+
+        except ValueError:
+            cls.log.warning("%r does not contain valid data for a UUID", path)
+            return None
```
opalmer: No arguments there, fixed in c8f4b11a54994905183b516b4423eb1277c45357.

pyfarm/agent/utility.py
opalmer: Done.  Configuration based defaults implemented in 56f46e4214701baa6b7ee883293f0515df9baf42.

pyfarm/agent/entrypoints/main.py
opalmer: Good points, I changed this to `--agent-id-file` in 56f46e4214701baa6b7ee883293f0515df9baf42 along side the default path changes.

guidow:
With the changes in pyfarm-core (https://github.com/pyfarm/pyfarm-core/pull/36) merged, I believe we are finally ready to merge this.

opalmer:
Thanks, sounds good to me.  I'll get this and the master changes merged.

PR #186, opened 2015-01-05
Do not convert agent_id to int in JobType.node()

guidow:
It is a UUID now, converting to int will fail.


opalmer:
Thanks, merging.  I didn't get a chance to test job types yet so I'm not surprised there's issues there still (especially since test coverage is lacking).

PR #187, opened 2015-01-05
Don't try to call .name on string object

guidow:


opalmer:
LGTM.  I'll merge this once the rerun of the one test failure completes.  I don't have access to a Windows system to test this on right now but the change makes sense and I'm guessing you opened this when you  hit a bug.

PR #188, opened 2015-01-05
Log a warning when a tempfile could not be deleted

guidow:
If a problem like this persists, it will clog up the temp directory.


opalmer:
Merging but I'm going to open up another pull request to improve on this some more.  This is a function that should probably always produce the same result once it's returned once.

PR #189, opened 2015-01-06
sysinfo.system Improvements

opalmer:
This is a follow-on to #188 that ensures `filesystem_is_case_sensitive` and `environment_is_case_sensitive` do not modify the file system or environment more often than they need to.


opalmer:
@guidow, PTAL.

guidow:
Looks good to me.

PR #191, opened 2015-01-07
Look for uuid.dat only in explicitly given path

guidow:
This path can be specified on the command line or in the config file,
but it has to be specified.  We no longer just try the various config
directories.


opalmer:
I just realized this pull request was open....I made similar changes in 56f46e4214701baa6b7ee883293f0515df9baf42 before I saw this, sorry about that.  You want me to roll that back or is there anything from here you want me to integrate into the change I made?

guidow:
Hm, no, I suppose your changes work, too, and they do the same thing in the end.

PR #192, opened 2015-01-09
Run control file

guidow:


pyfarm/agent/entrypoints/main.py
```diff
@@ -621,6 +630,27 @@ def remove_pid_file_on_exit():
 
         logger.info("pid: %s" % pid)
 
+        if not isfile(config["run_control_file"]):
+            directory = dirname(config["run_control_file"])
+            if not isdir(directory):
```
opalmer: Rather than checking with isdir, catch OSError\IOError and check errno against EEXIST.

pyfarm/agent/entrypoints/main.py
```diff
@@ -621,6 +630,27 @@ def remove_pid_file_on_exit():
...
+                    os.makedirs(directory)
+                except OSError:  # pragma: no cover
+                    logger.error(
+                        "Failed to create parent directory for %s",
+                        config["run_control_file"])
+                    raise
+                else:
+                    logger.debug("Created directory %s", directory)
+            try:
+                control_file = open(config["run_control_file"], "a").close()
```
opalmer:
```python
with open(path, "w"):
    pass
```

pyfarm/agent/entrypoints/main.py
```diff
@@ -621,6 +630,27 @@ def remove_pid_file_on_exit():
...
         logger.info("pid: %s" % pid)
 
+        if not isfile(config["run_control_file"]):
+            directory = dirname(config["run_control_file"])
+            if not isdir(directory):
+                try:
+                    os.makedirs(directory)
+                except OSError:  # pragma: no cover
+                    logger.error(
+                        "Failed to create parent directory for %s",
```
opalmer: Include the exception in the error message so we can figure out what went wrong more easily.

pyfarm/agent/etc/agent.yml
```diff
@@ -63,6 +70,7 @@ agent_api_port: 50000
 # The location where the the agent should save its own
 # logging output to.
 agent_log: $agent_logs_root/agent.log
+agent_log: $agent_logs_root/supervisor.log
```
opalmer: Hmm.  Could you explain why this should be supervisor.log instead of agent.log?  The agent itself could be writing to this location and while the supervisor could have a log too I assume we'd want either a combined log file or a separate path for each.

In either case, the log file is not currently implemented....need to fix that...

pyfarm/agent/http/api/state.py
```diff
@@ -46,6 +53,15 @@ def post(self, **kwargs):
...
         data = kwargs["data"]
         agent = config["agent"]
+
+        try:
+            os.remove(config["run_control_file"])
+        except (WindowsError, OSError, IOError) as e:
+            if e.errno != ENOENT:
+                logger.error("Could not delete run control file %s: %s: %s",
+                             config["run_control_file"],
+                             type(e).__name__, e)
```
opalmer: Probably should add a call to `atexit.register` here to try and clean the file when Python exits.

pyfarm/agent/http/api/state.py
```diff
@@ -67,6 +83,43 @@ def finished(_, finish_request):
...
+    isLeaf = False
+
+    def post(self, **kwargs):
+        request = kwargs["request"]
+        agent = config["agent"]
+
+        #Ensure the run control file exists
+        if not isfile(config["run_control_file"]):
+            directory = dirname(config["run_control_file"])
+            if not isdir(directory):
```
opalmer: same as above, I'd try os.makedirs and check the error

pyfarm/agent/http/api/state.py
```diff
@@ -67,6 +83,43 @@ def finished(_, finish_request):
...
+                    os.makedirs(directory)
+                except OSError:  # pragma: no cover
+                    logger.error(
+                        "Failed to create parent directory for %s",
+                        config["run_control_file"])
+                    raise
+                else:
+                    logger.debug("Created directory %s", directory)
+            try:
+                control_file = open(config["run_control_file"], "a").close()
```
opalmer: same as above, I think using the context manager is cleaner

pyfarm/agent/service.py
```diff
@@ -443,6 +444,14 @@ def wait_on_stopped():
...
 
     def sigint_handler(self, *_):
+        try:
+            os.remove(config["run_control_file"])
+        except (WindowsError, OSError, IOError) as e:
+            if e.errno != ENOENT:
+                logger.error("Could not delete run control file %s: %s: %s",
+                             config["run_control_file"],
+                             type(e).__name__, e)
+
```
opalmer: same as above, consider registering an exit handler with `atexit.register`

tests/test_agent/test_http_api_state.py
```diff
@@ -38,6 +38,10 @@ class TestStop(BaseAPITestCase):
     URI = "/stop"
     CLASS = Stop
 
+    def prepare_config(self):
+        super(TestStop, self).prepare_config()
+        config.update(run_control_file="/tmp/pyfarm/agent/should_be_running")
```
opalmer: Use `self.create_file()` in the tests.  It's cross-platform and will manage the file handle.

pyfarm/agent/etc/agent.yml
guidow: The second line was supposed to start with ```supervisor_log```.

> I assume we'd want either a combined log file or a separate path for each.

Two processes writing to the same file simultaneously could be hairy, and AFAIK, won't even work on Windows.

pyfarm/agent/http/api/state.py
guidow: I could do that, I suppose, but if we cannot delete the file at this point, at is very unlikely we will be able to delete it on exit.

pyfarm/agent/service.py
guidow: As said above, I don't see why it would work any better as an exit handler than right here.

tests/test_agent/test_http_api_state.py
guidow: I'm not creating the file at that point, just setting an entry in the configuration.

pyfarm/agent/http/api/state.py
opalmer: At least on Windows I've seen the behavior several times where os.remove doesn't work because Windows still considers the file locked for some reason.  Usually, calling os.remove as the interpreter exits will clean it up.

pyfarm/agent/etc/agent.yml
opalmer: > AFAIK, won't even work on Windows.

It can....but not under normal circumstances that I'd call normal but I'd thought I'd ask to be sure.  Thanks for clearing this up.

guidow:
Any more objections to merging this?

opalmer:
Nope none, merging.  Didn't mean to leave this one sitting out for so long, sorry my bad.

PR #193, opened 2015-01-09
Switch to @inlineCallbacks

opalmer:
**This branch is a work in progress**

Twisted's `@inlineCallbacks` offer a cleaner method of writing code.  This PR should not change how the code functions.  It should instead cleanup the code we already have, making it easier to code and debug.

Instead of code like this:

```python
def foo():
    def success(result):
        print result

    def failure(failure):
        print failure

    defer = deferLater(reactor, 1, lambda: None)
    defer.addCallback(success)
    defer.addErrback(failure)
    return defer
```

You can rewrite it like this and handle the error in a more liner flow while still running asynchronously:

```python
@inlineCallbacks
def foo():
    try:
         result = yield deferLater(reactor, 1, lambda: None)
    except Exception as failure:  # errback
        print failure
    else:  # callback
        returnValue(result)
```


opalmer:
@guidow, I'd like you to a look at the work done so far on this.  There's still quite a ways to go but I think getting some input now would be good before I dive too deeply into the updating API and job types with inlineCallbacks.

I'm not planning on changing the external representation of either of those.  So in the end everything should be functionally equivalent from an external interaction perspective (which is not to say we shouldn't test it to be sure).

guidow:
Sorry, I didn't get to take a good look today. Will try again tomorrow.

opalmer:
@guidow, did you get a chance to have a look?

opalmer:
I'm splitting up some of these changes into smaller reviews where possible.  That should be a little easier to review probably (this is still a work in progress though).

guidow:
Again, sorry I haven't managed to have look yet – I caught some sickness last week and wasn't able to work...

guidow:
> I'm splitting up some of these changes into smaller reviews where possible. Reviewing those changes first would probably be a good idea since they're smaller and would reduce the size of this review.

Thanks for this.

To clarify: Does that make this pull request redundant?

opalmer:
> Again, sorry I haven't managed to have look yet – I caught some sickness last week and wasn't able to work...

No worries, hope you're feeling better.

> To clarify: Does that make this pull request redundant?

Yes, most likely so I'm going to close this out in favor of smaller changes that can be merged without breaking the existing code base.  Normally breakage is easy to avoid but for this kind of change smaller is definitely going to be easier to manage....

PR #194, opened 2015-01-11
Fix logged pid in startup

opalmer:
This is a simple pull request to fix the rendering of the pid in a log message when the agent starts up.


opalmer:
Merging directly once the tests pass since this is a trivial change.

PR #195, opened 2015-01-12
Replace httpbin in tests

opalmer:
Many of our recent test failure are because `http[s]://httpbin.org` has been down or slow to respond, especially over SSL.  This change changes the url we're testing against to point at one of pyfarm's web servers instead.


opalmer:
Merging because this is a non-functional change.  The only impact is an improvement to test reliability.

PR #196, opened 2015-01-13
Fix test when the test machine actually has a GPU

guidow:
The broken nature of thhis test has never surfaced until now, because
the Travis instances don't have GPUs or cannot list them with lspci.


opalmer:
lgtm, thanks.

PR #197, opened 2015-01-14
Accept args in lambda for jobtype download errback

guidow:
This fixes a bug which would leave the agent in an usable state if it
could not reach the master when trying to download a jobtype file.


opalmer:
lgtm, thanks.

PR #198, opened 2015-01-21
remove unused http.api.log since we have tasklogs

opalmer:
Since we're not using this module and probably never will be, we should remove it.


opalmer:
@guidow, PTAL.  I don't see us having a reason to keep this module around.

guidow:
Sure. Removing unused code is always a good idea.

PR #200, opened 2015-01-21
Sigint handler fix

opalmer:
This fixes a bug introduced in #192.  It's my fault for not catching this in the review though #145 probably would have caught this (my IDE noticed it when I tried merging the master into my working branch).  Sorry about that...


opalmer:
@guidow, PTAL

guidow:
Looks good to me.

PR #201, opened 2015-02-02
Improved _remove_directories()

opalmer:
This change improves on _remove_directories() inside internals.py a little by:

* Using one thread to cleanup multiple directories
* Only making one attempt to remove a directory, instead of several in a very short period of time, before scheduling a retry on exit.


opalmer:
@guidow, PTAL

guidow:
Looks good.

PR #202, opened 2015-02-02
Inline Callbacks: Service

opalmer:
This change is a portion of the changes in #193.  Tests won't pass however because this changes the client program which breaks the job types.


opalmer:
@guidow, PTAL

pyfarm/agent/service.py
```diff
@@ -84,18 +87,19 @@ def __init__(self):
         self.services = {}
         self.register_shutdown_events = False
         self.last_free_ram_post = time.time()
-        self.shutting_down = False
```
guidow: This is used in at least one more place (assign.py). Simply removing this breaks assign.

pyfarm/agent/service.py
```diff
@@ -84,18 +87,19 @@ def __init__(self):
...
-        # reannounce_client is set when the agent id is
-        # first set. reannounce_client_instance ensures
-        # that once we start the announcement process we
-        # won't try another until we're finished
-        self.reannounce_client_request = None
-
-        # Setup scheduled tasks
-        self.scheduled_tasks = ScheduledTaskManager()
-        self.scheduled_tasks.register(
-            self.reannounce, config["agent_master_reannounce"])
```
guidow: Isn't this effectively disabling the reannounce?

guidow:
The code looks mostly okay to me (though obviously there are still failing tests), however I'm still trying to wrap my head around exactly how these inline callbacks are supposed to work...

pyfarm/agent/service.py
opalmer: I'll be sure to readd this, it shouldn't have been removed.

opalmer:
> however I'm still trying to wrap my head around exactly how these inline callbacks are supposed to work...

They're great...once you get your head around them (like most things in Twisted).  Some of the primary reasons I'm making this move is to make the internal code flow is more linear, easier to debug and less likely to get hung up by something like a missed `.addErrback`.  I mainly got used to inlineCallbacks by using, them but if I had to suggest some online reading I'd go with these to start with:

http://hackedbellini.org/development/writing-asynchronous-python-code-with-twisted-using-inlinecallbacks/
http://krondo.com/?p=2441
http://twisted.readthedocs.org/en/latest/core/howto/defer-intro.html#inline-callbacks-using-yield (the whole page is useful, but the inlineCallbacks stuff starts here)

pyfarm/agent/service.py
opalmer: It would be replaced with something new that's smaller but does the same thing:

https://github.com/pyfarm/pyfarm-agent/commit/19ccbdd8d380f9339d33dd747e63101c600b37a5

This code here is probably something I'll make into it's own change....

opalmer:
@guidow, I'm going to leave this review open but focus on other changes that probably need to come before this one does.

PR #205, opened 2015-02-02
Implement utility method for cleaning up temp dirs

guidow:
To be used by jobtype implementors

### DIFF
--- a/pyfarm/jobtypes/core/internals.py
+++ b/pyfarm/jobtypes/core/internals.py
@@ -28,6 +28,7 @@ import os
 import sys
 import shutil
 import tempfile
+import time
 from collections import namedtuple
 from errno import EEXIST, ENOENT
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


pyfarm/jobtypes/core/internals.py
```diff
@@ -792,6 +797,39 @@ def rmdir(path, on_exit_handler=True):
         for directory in self._tempdirs:
             reactor.callInThread(rmdir, directory)
 
+    def _ensure_free_space_in_temp_dir(self, tempdir, space):
```
opalmer: What do you think of adding an optional `minimum_age` keyword?  The idea being that if provided a timedelta object only files older than `minimum_age` would be cleaned up and all others would be ignored.  I realize this should cleanup based on atime but in some situations, such as when a render runs on an artist's machine, I'm not sure that alone will be sufficient.

opalmer:
Looks good to me, one comment to consider first though I think.

opalmer:
lgtm, merging.

PR #206, opened 2015-02-03
Improved HTTP Client Setup and Validation

opalmer:
This change improves on some validation of the incoming request and also does a bit of reorganization of `request()`.  This change is in preparation for another change which will implement the new http client functionality needed for inlineCallbacks.

### DIFF
--- a/pyfarm/agent/http/core/client.py
+++ b/pyfarm/agent/http/core/client.py
@@ -331,17 +331,8 @@ def request(method, url, **kwargs):
     """
     assert isinstance(url, STRING_TYPES)
 
-    # Make sure it's a method we support.
-    if method not in HTTP_METHODS:
-        raise NotImplementedError(
-            "This function only supports these http "
-            "methods: %s" % ", ".join(HTTP_METHODS))
-
     # We only support http[s]
     parsed_url = urlparse(url)
-    if parsed_url.scheme not in HTTP_SCHEMES:
-      raise NotImplementedError("Only http or https is supported.")
-
     if not parsed_url.hostname:
         raise NotImplementedError("No hostname present in url")
 
--- a/pyfarm/agent/http/core/client.py
+++ b/pyfarm/agent/http/core/client.py
@@ -381,7 +381,19 @@ def request(method, url, **kwargs):
             raise NotImplementedError(
                 "cannot handle header values with type %s" % type(value))
 
+    # Handle request data
     data = kwargs.pop("data", NOTSET)
+    if isinstance(data, dict):
+        data = json_safe(data)
+
+    if (data is not NOTSET and
+            headers["Content-Type"] == ["application/json"]):
+        data = json.dumps(data)
+
+    elif data is not NOTSET:
+        raise NotImplementedError(
+            "Don't know how to dump data for %s" % headers["Content-Type"])
+
     callback = kwargs.pop("callback", None)
     errback = kwargs.pop("errback", log.err)
     response_class = kwargs.pop("response_class", Response)
@@ -404,18 +416,6 @@ def request(method, url, **kwargs):
             response_class(deferred, response, original_request))
 
         return deferred
-
-    # prepare to send the data
-    if isinstance(data, dict):
-        data = json_safe(data)
-
-    if data is not NOTSET and \
-                    headers["Content-Type"] == ["application/json"]:
-        data = json.dumps(data)
-
-    elif data is not NOTSET:
-        raise NotImplementedError(
-            "Don't know how to dump data for %s" % headers["Content-Type"])
 
     # prepare keyword arguments
     kwargs.update(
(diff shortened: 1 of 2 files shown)


opalmer:
@guidow PTAL.  This change just improves on the validation leading up to building a request and is a precursor to #208.

pyfarm/agent/http/core/client.py
```diff
@@ -304,35 +323,56 @@ def request(method, url, **kwargs):
...
+            "methods: %s" % ", ".join(HTTP_METHODS))
+
+    # We only support http[s]
+    parsed_url = urlparse(url)
+    if parsed_url.scheme not in HTTP_SCHEMES:
+      raise NotImplementedError("Only http or https is supported.")
+
+    # When using HTTPs there are some specific modules that are
+    # required but not specified as required in the setup.py file.
+    if parsed_url.scheme == "https":
```
guidow: Is this check really necessary? If a client were to connect with SSL when PyOpenSSL is not available, this part should not even be reached, as far as I can tell.

pyfarm/agent/http/core/client.py
```diff
@@ -304,35 +323,56 @@ def request(method, url, **kwargs):
...
+    # Twisted requires lists for header values
     for header, value in headers.items():
         if isinstance(value, STRING_TYPES):
             headers[header] = [value]
 
-        # for our purposes we should not expect headers with more
+        # For our purposes we should not expect headers with more
         # than one value for now
         elif isinstance(value, (list, tuple, set)):
             assert len(value) == 1
```
guidow: More than one instance of a header key is technically allowed by HTTP, although unusual in practice. Why do we need to explicitly catch that case here?

pyfarm/agent/http/core/client.py
```diff
@@ -304,35 +323,56 @@ def request(method, url, **kwargs):
...
+    # Make sure it's a method we support.
+    if method not in HTTP_METHODS:
+        raise NotImplementedError(
+            "This function only supports these http "
+            "methods: %s" % ", ".join(HTTP_METHODS))
+
+    # We only support http[s]
+    parsed_url = urlparse(url)
+    if parsed_url.scheme not in HTTP_SCHEMES:
+      raise NotImplementedError("Only http or https is supported.")
```
guidow: Why do we need to catch that here manually? If treq does not support that scheme, it will raise an exception eventually anyway.

pyfarm/agent/http/core/client.py
```diff
@@ -304,35 +323,56 @@ def request(method, url, **kwargs):
...
+
+    # When using HTTPs there are some specific modules that are
+    # required but not specified as required in the setup.py file.
+    if parsed_url.scheme == "https":
+        if PyOpenSSL is NotImplemented and ssl is NotImplemented:
+            raise NotImplementedError(
+                "HTTPs requires the ssl or PyOpenSSL modules")
```
guidow: Why do we need to catch that case here manually? If treq does not support SSL, it will raise an exception eventually anyway.

pyfarm/agent/http/core/client.py
opalmer: Technically, we shouldn't need to catch it.  None of our internal code does this however which is why it's here now.  That said, I'd think it's probably better if we align more so what HTTP can do rather than what we're doing internally (esp. if someone else decides to use this in the future)

pyfarm/agent/http/core/client.py
opalmer: The error treq *used* to raise would tell you nothing about what the real problem was and you'd have to investigate because it was similar to other errors.  That's not the case anymore however it seems, I just checked and now we get this:
```
  File "twisted/web/client.py", line 1560, in request
    endpoint = self._getEndpoint(parsedURI)
  File "twisted/web/client.py", line 1544, in _getEndpoint
    return self._endpointFactory.endpointForURI(uri)
  File "twisted/web/client.py", line 1420, in endpointForURI
    tlsPolicy = self._policyForHTTPS.creatorForNetloc(uri.host, uri.port)
  File "twisted/web/client.py", line 797, in raiseNotImplemented
    raise NotImplementedError("SSL support unavailable")
NotImplementedError: SSL support unavailable
```

So I'll remove this check (sorry didn't do this before, hadn't updated my treq client in a while)

pyfarm/agent/http/core/client.py
opalmer: Same answer as below, the error used to be unhelpful in telling you what was wrong.  But also like below, I think more recent versions of Twisted have improved on this:

```
  File "twisted/web/client.py", line 1560, in request
    endpoint = self._getEndpoint(parsedURI)
  File "twisted/web/client.py", line 1544, in _getEndpoint
    return self._endpointFactory.endpointForURI(uri)
  File "twisted/web/client.py", line 1424, in endpointForURI
    raise SchemeNotSupported("Unsupported scheme: %r" % (uri.scheme,))
twisted.web.error.SchemeNotSupported: Unsupported scheme: 'foo'
```

PR #207, opened 2015-02-03
Silence Configuration Logging during setUp()

opalmer:
Currently the config logger is very noisy during the tests making it difficult to quickly find issues.  This change fixes that issue by implementing `.disabled` on our logger class (which we were using but it didn't in fact do anything).

For a comparison, here's [before](https://s3.amazonaws.com/archive.travis-ci.org/jobs/49269512/log.txt) and here's [after](https://s3.amazonaws.com/archive.travis-ci.org/jobs/49270476/log.txt) this change.  Now the only config changes that are logged are those that take place within the test itself.

### DIFF
--- a/pyfarm/agent/logger/python.py
+++ b/pyfarm/agent/logger/python.py
@@ -40,30 +40,37 @@ class Logger(object):
     """
     def __init__(self, name):
         self.name = name
+        self.disabled = False
 
     def debug(self, message, *args):
-        msg(message, args=args, system=self.name,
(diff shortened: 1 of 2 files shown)


opalmer:
@guidow, PTAL

guidow:
Looks good.

Good idea anyway, the logs could often be far too large to be useful.

opalmer:
Yeah that's exactly why I made this change pretty much, Travis was having issues displaying the logs sometimes.

PR #208, opened 2015-02-03
Direct HTTP Client (treq) Request Support

opalmer:
This change adds support to use the treq client as directly as possible and is a continuation of #206 (which should merge first).  This change is also a prerequisite for the inlineCallback work that's being done in other branches but we should be able to merge this without breaking the agent as it stands now.

### DIFF
--- a/pyfarm/agent/http/core/client.py
+++ b/pyfarm/agent/http/core/client.py
@@ -330,6 +330,7 @@ def request(method, url, **kwargs):
         constructing data to an api.
     """
     assert isinstance(url, STRING_TYPES)
+    direct = kwargs.pop("direct", False)
 
     # Make sure it's a method we support.
     if method not in HTTP_METHODS:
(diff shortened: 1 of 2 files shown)


opalmer:
@guidow, PTAL after reviewing #206 since this is just built on top of that change.  I will be writing some tests for the new code before merging this but functionally this shouldn't change between now and when the tests are finished.

guidow:
Could you please explain what this new direct option does and, more importantly, how it's supposed to improve things?

opalmer:
The direct option removes the abstraction around the way we currently handle the callback/errbacks and just returns the treq requests for the caller to handle directly.  For the current code base, this change will not do anything because I didn't change the get/post/etc partials.  For newer changes however, such as the inlineCallbacks I'm working on, this enables finer grained control over when and where we handle the results of the request.  This change also reduces the memory footprint a little for direct requests because we're not longer copying around the original request (which the caller should handle).

guidow:
I have looked over the code and tried the branch in practice for a short test run, and did not find any real problems.

opalmer:
Thanks, I'm going to go ahead and merge this and #206 then so I can get back to work on the inlineCallbacks changes.

Issue #209, opened 2015-02-03
Assertion error in _remove_directories()


guidow:
I found the following error in the agent output today


    2015-02-03 21:46:42 DEBUG    - twisted         - Unhandled error in Deferred:
    2015-02-03 21:46:42 ERROR    - root            - Unhandled Error
    Traceback (most recent call last):                                                                                                                                                                                              
    Failure: twisted.internet.error.ProcessTerminated: A process has ended with a probable error condition: process ended by signal 9.                                                                                              
    Traceback (most recent call last):                                                                                                                                                                                              
      File "/home/guido/git/pyfarm/pyfarm-agent/pyfarm/agent/entrypoints/main.py", line 659, in start                                                                                                                               
        os.makedirs(directory)                                                                                                                                                                                                      
      File "/var/tmp/virtualenv/pyfarm-agent-python2/lib64/python2.7/os.py", line 157, in makedirs                                                                                                                                  
        mkdir(name, mode)                                                                                                                                                                                                           
    OSError: [Errno 17] File exists: '/tmp/pyfarm/agent'                                                                                                                                                                            
    2015-02-03 21:46:42 ERROR    - root            - Unhandled Error
    Traceback (most recent call last):                                                                                                                                                                                              
      File "/usr/lib64/python2.7/threading.py", line 783, in __bootstrap                                                                                                                                                            
        self.__bootstrap_inner()                                                                                                                                                                                                    
      File "/usr/lib64/python2.7/threading.py", line 810, in __bootstrap_inner                                                                                                                                                      
        self.run()                                                                                                                                                                                                                  
      File "/usr/lib64/python2.7/threading.py", line 763, in run                                                                                                                                                                    
        self.__target(*self.__args, **self.__kwargs)                                                                                                                                                                                
    --- <exception caught here> ---                                                                                                                                                                                                 
      File "/var/tmp/virtualenv/pyfarm-agent-python2/lib/python2.7/site-packages/Twisted-14.0.2-py2.7-linux-x86_64.egg/twisted/python/threadpool.py", line 196, in _worker                                                          
        result = context.call(ctx, function, *args, **kwargs)                                                                                                                                                                       
      File "/var/tmp/virtualenv/pyfarm-agent-python2/lib/python2.7/site-packages/Twisted-14.0.2-py2.7-linux-x86_64.egg/twisted/python/context.py", line 118, in callWithContext                                                     
[...]

PR #210, opened 2015-02-04
Fix type check in _remove_directories

opalmer:
This should fix the assertion error in #209.  The problem is caused by `self._tempdirs` being a set object originally.


opalmer:
@guidow, PTAL.  This should fix #209.

guidow:
Looks good and seems to work. I see no more AssertionErrors with this one.

PR #213, opened 2015-02-06
Synchronize Agent.shutting_down with the config

opalmer:
Right now setting `shutting_down` on the agent will not necessarily communicate to other parts of the project, such as job types or other code which does not have easy access to the agent instance, that the agent is shutting down.  This change fixes this by turning `shutting_down` into a property that will return a result from the config when `self.shutting_down` is requested and set the value in the config with `self.shutting_down = <value>`


opalmer:
@guidow, PTAL

opalmer:
@guidow, any comments?

guidow:
Looks fine, seems to work.

PR #214, opened 2015-02-06
Replace ScheduledTaskManager with Simplified Method

opalmer:
This change replaces `ScheduledTaskManager` with a simpler callable called `schedule_call`.  Functionally this should operate the same as it did before.


opalmer:
@guidow, PTAL.  I'm going to be writing some tests for this tomorrow but code wise it should be ready for a review.

opalmer:
Updated, tests are included now.

opalmer:
@guidow, any comments?

opalmer:
@guidow, any comments?

pyfarm/agent/service.py
```diff
@@ -118,6 +115,60 @@ def agents_endpoint(cls):
         """
         return config["master_api"] + "/agents/"
 
+    @inlineCallbacks
+    def schedule_call(self, delay, function, *args, **kwargs):
+        """
+        Schedules ``function`` to be run after ``delay`` has elapsed.
```
guidow: Maybe slightly misleading – by default, this will call ```function``` repeatedly and indefinitely every ```delay```. The behaviour described here can only be achieved by explicitly setting ```repeat``` to 1.

pyfarm/agent/service.py
```diff
@@ -118,6 +115,60 @@ def agents_endpoint(cls):
...
+            Number of seconds to delay calling ``function``
+
+        :param function:
+            A callable function to run
+
+        :param args:
+            Arguments to pass into ``function``
+
+        :param kwargs:
+            Keywords to pass into ``function``
```
guidow: These two arguments should be differently named. As it stands now, scheduled functions cannot have parameters with the same name as any parameter of ```schedule_call```.

pyfarm/agent/service.py
```diff
@@ -118,6 +115,60 @@ def agents_endpoint(cls):
...
+            )
+            yield deferLater(self.reactor, 0, function, *args, **kwargs)
+
+        if repeat is None or repeat > 0:
+            repeat -= 1
+            kwargs.update(repeat=repeat)
+            yield deferLater(
+                self.reactor, delay, self.schedule_call, delay, function,
+                *args, **kwargs
+            )
```
guidow: This strategy of running a task periodically will introduce drift over time, possibly a lot. To avoid that drift, we would have to schedule every call for a specific point in time and then schedule the call after that to run at that point plus ```delay``` seconds.

This is probably not that important for what we are doing, however a user of this function might assume that if they schedule a function to be called once every minute, that it will be called 60 times in an hour and 1440 times in a day. The delay means that, by the time it is run for the 60th time, more than an hour will have passed, possibly a lot more, depending on various circumstances.

pyfarm/agent/service.py
```diff
@@ -774,4 +823,5 @@ def callback_agent_id_set(
                 "`%s` was %s, adding system event trigger for shutdown",
                 key, change_type)
 
-            self.scheduled_tasks.start()
+            self.schedule_call(
+                config["agent_master_reannounce"], self.reannounce)
```
guidow: This line is never reached in my tests, probably because ```agent_id``` is modified but not created on startup. Not a problem with this specific pull request, though.

pyfarm/agent/service.py
```diff
@@ -118,6 +115,60 @@ def agents_endpoint(cls):
...
+
+        if now:
+            svclog.debug(
+                "Executing task %s(*%r, **%r).  Next execution in %s seconds.",
+                function.__name__, args, kwargs, delay
+            )
+            yield deferLater(self.reactor, 0, function, *args, **kwargs)
+
+        if repeat is None or repeat > 0:
+            repeat -= 1
```
guidow: Breaks when repeat is None, which it is by default.

pyfarm/agent/service.py
```diff
@@ -77,6 +77,8 @@ class Agent(object):
     itself with the master, starting the periodic task manager,
     and handling shutdown conditions.
     """
+    reactor = reactor
```
guidow: Why is this necessary?

pyfarm/agent/service.py
opalmer: I'll rename this then to something like 'repeating_call' or something similar.  I carried over the old name initially but you're right that it does not make as much sense given the context.

pyfarm/agent/service.py
opalmer: Yep, you're right it's never reached in the tests.  It's probably the same issue I encountered in the tests for #215, which I discovered after updating the tests here, so I'll see if I can fix the tests so we do reach this line.

pyfarm/agent/service.py
opalmer: Same reason for why #215 has:

```
It's not anymore, removing. Anytime you see something like this it's usually the reactor can be replaced in testing. In the case of this change I have not had a reason to use it for testing so far...
```
I'm removing the change in this PR for the same reason.

pyfarm/agent/service.py
opalmer: Fixing.

pyfarm/agent/service.py
opalmer: Fixed.

pyfarm/agent/service.py
opalmer: I'll add a note in the doc string about this.  I agree that there could be a decent amount of drift here due to how the operates but I also agree that it's not the goal to deal in absolute times here either.  But, that should be more clear in the documentation string.

opalmer:
PTAL, I think I've addressed the issues you've brought up.  I've also removed `inlineCallbacks` for scheduled_call due to what appear to be some recursion issues on exit.  Commit c5381025cb389593e6feecc8e0b1afcbe3f5c167 cover this in more detail and in the end it's not 100% necessary here either it turned out.

guidow:
>  I've also removed inlineCallbacks for scheduled_call due to what appeared to be some recursion issues on exit. Commit c538102 covers this in more detail and in the end inlineCallbacks are not not 100% necessary here either it turned out.

Thinking about this some more, this might be better anyway. The original could have produced an ever growing chain of mostly dormant coroutines all waiting for the next one in line to finish, which never happens.

Have you tried removing just last ```yield```?

opalmer:
> Have you tried removing just last yield? I think if you do this, the original coroutine will reach the end and be removed from memory without having to wait for the next one.

I had not actually but I just tried it and it does in fact fix the problem (which is good to know for the future, good catch).  

I still think removing inlineCallbacks all together though from this one is probably the right call for this one.  With that I think this one is ready to merge unless you think there's something else to still address on this one.

PR #215, opened 2015-02-09
Convert post_agent_to_master to inlineCallbacks

opalmer:
Similar to other recent branches the focus of this change is to convert a part of `service.py` to use inlineCallbacks.  In this case, the part which is performing a POST request to register the agent with the master.

### DIFF
--- a/pyfarm/agent/service.py
+++ b/pyfarm/agent/service.py
@@ -631,7 +631,6 @@ class Agent(object):
                     "responded with was %r.  Sorry, but we cannot retry this "
                     "request as it's an issue with the agent's request.",
                     url, response.code, text)
-                returnValue(None)
 
             else:
                 data = yield treq.json_content(response)
--- a/pyfarm/agent/service.py
+++ b/pyfarm/agent/service.py
@@ -591,7 +591,7 @@ class Agent(object):
                 svclog.error(
                     "Failed to POST agent to master, the connection was "
                     "refused. Retrying in %s seconds")
-            else:
+            else:  # pragma: no cover
                 svclog.error(
                     "Unhandled error when trying to POST the agent to the "
                     "master. The error was %s.", failure)
--- a/pyfarm/agent/service.py
+++ b/pyfarm/agent/service.py
@@ -568,31 +568,6 @@ class Agent(object):
 
         return finished
 
-    def errback_post_agent_to_master(self, failure):
-        """
-        Called when there's a failure trying to post the agent to the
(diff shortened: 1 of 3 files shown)


opalmer:
@guidow, PTAL.   Will start working on some tests for this soon but the general code is here.

pyfarm/agent/service.py
```diff
@@ -564,88 +568,84 @@ def error_while_posting(failure):
...
-        # Master is up but is rejecting our request because there's something
-        # wrong with it.  Do not retry the request.
-        elif response.code >= 400:
-            svclog.error(
-                "%s accepted our POST request but responded with code %s "
-                "which is a client side error.  The message the server "
-                "responded with was %r.  Sorry, but we cannot retry this "
-                "request as it's an issue with the agent's request.",
-                self.agents_endpoint(), response.code, response.data())
+                svclog.warning("Not retrying POST to master, shutting down.")
```
guidow: Misleading log message – the agent is not shutting down at this point.

pyfarm/agent/service.py
```diff
@@ -77,6 +79,8 @@ class Agent(object):
     itself with the master, starting the periodic task manager,
     and handling shutdown conditions.
     """
+    reactor = reactor
```
guidow: Why is this necessary?

guidow:
I have looked over the code and tried the branch in practice. It works, and I could not find any real problems.

guidow:
Works in principle, but has a few bugs left.

pyfarm/agent/service.py
opalmer: It's not anymore, removing.  Anytime you see something like this it's usually the reactor can be replaced in testing.  In the case of this change I have not had a reason to use it for testing so far...

opalmer:
Thanks for the review.  Still finishing the tests for this so I wouldn't be surprised if there's a few issues lingering.  Are the bugs you're seeing from reviewing the code or through direct testing (trying to make sure I don't miss anything in the tests I'm writing).

guidow:
I'm sorry, the comment with "has a few bugs left" should have gone to the other PR...

opalmer:
Ah ok that makes sense, thanks for clarifying.  I'm going to continue adding the remaining tests for this then and once that's complete I'll go ahead and merge, assuming there are not any major changes to the code, unless you have objections.

opalmer:
Going to go ahead and merge this since there are not any functional changes since the last review.

PR #216, opened 2015-02-12
Internal message logging

guidow:


opalmer:
Looks ok to me.  One question I have though is why `_log` instead of `log`?  I thought we were going to reserve 'private' methods for internal use or things a job type implementor generally wouldn't use.

guidow:
Hm, my understanding was that, as far as the jobtype API is concerned, leading underscore means the jobtype implementor is not supposed to override that method.

PR #217, opened 2015-02-13
Make Retry Delay Fully Configurable

opalmer:
While working on the tests for #215 it became apparent that it was not possible to globally configure the minimum for `http_retry_delay`.  This change fixes that and also fixes an existing test too.


opalmer:
@guidow, PTAL.

guidow:
This code returns the effective delay as ```return max(minimum, delay)```. This is problematic, because in practice ```minimum``` will be returned more often than any other value, skewing the randomness. This problem gets more pronounced with higher values for minimum, and as soon as minimum is equal to or larger than ```agent_http_retry_delay```, the randomness will be gone completely.

We are using this random delay to avoid livelock situations. If the master ends up refusing some requests because of overload because too many agents are attempting to make requests at the same time, it won't help us if they all wait the exact same time and try their request again at the same point. If an administrator explicitly and deliberately raises the minimum for delays, that probably means he is experiencing overload problems already – which makes it even more important that this works correctly.

My suggestion is:

- Rename ```agent_http_retry_delay``` to ```agent_http_retry_delay_factor```
- Rename ```minimum``` to ```offset```
- Rewrite this function as ```return offset + random() * agent_http_retry_delay_factor```

It really doesn't need to be any more complex than this.

opalmer:
@guidow, PTAL

opalmer:
@guidow, do the latest changes cover your comments?

guidow:
Looks okay to me now.

guidow:
The old, now no longer existing config variable ```agent_http_retry_delay``` is still being referenced in the following places:

- pyfarm/agent/http/system.py, lines 133 and 136
- pyfarm/agent/testutil.py, line 390
- pyfarm/agent/entrypoints/main.py, line 376

The agent won't start because of this.

opalmer:
Wow my bad, I should have grepped around for it before requesting another look.  Sorry about that.  I've fixed those locations and also added some better information in the yaml file too.

guidow:
Looks good now, and the code works again. Retry values appear to be random enough.

PR #218, opened 2015-02-14
Convert post_free_ram to inlineCallbacks

opalmer:
Similar to other recent branches the focus of this change is to convert a part of `service.py` to use inlineCallbacks.  In this case, the part which is performing a POST request update the amount of free ram on the agent to the master.


opalmer:
@guidow PTAL, this is ready for review/merging.

guidow:
The code looks okay, but in practice does not seem to get called at all in my tests.

I suppose what's happening here is that automatic periodic updates of free ram have been broken for some time, and we just havent't noticed yet.

opalmer:
That's likely because the free ram post is not currently scheduled in the job type like it needs to be at the moment.  When the agent self post it includes the amount of free ram, post_free_ram_to_master itself is meant to be called in between instance of a full post and at specific points when the job type runs.  Right now it's not doing that though (but will be in the future)

guidow:
Why not just do it as part of the periodic reannounce? As far as I can see, this would be the simplest solution to the problem (provided the periodic reannounce actual works...) and do everything we need.

opalmer:
Ok, no problem with that.  The original reason I didn't do that is because it's that the periodic reannounce may not be enough to keep up with changing resources.  That said....we can always add this back in later if that turns out to be the case.

PR #219, opened 2015-02-17
Convert post_cpu_count to inlineCallbacks

opalmer:
Similar to other recent branches the focus of this change is to convert a part of service.py to use inlineCallbacks. In this case, the part which is performing a POST request update the number of cpus on the agent to the master.


opalmer:
@guidow, PTAL

guidow:
Why do we even have a function post_cpu_count? Is this value likely to change at runtime? I would think not.

opalmer:
Usually not no, but if someone updates the config we'd have to trigger the update.  That said I think idea I had in #218 about less specific update functions would probably be a better idea than this PR.

PR #220, opened 2015-02-19
Remove CPU/RAM Update Methods

opalmer:
As discussed in #218 and #219 this change removes the code which POSTs CPU and RAM values to the master.  This code is currently unused and is being removed in favor os the reannouce functionality.

### DIFF
--- a/pyfarm/agent/service.py
+++ b/pyfarm/agent/service.py
@@ -49,7 +49,6 @@ import treq
 from ntplib import NTPClient
 from twisted.internet import reactor
 from twisted.internet.defer import Deferred, inlineCallbacks, returnValue
-from twisted.internet.task import deferLater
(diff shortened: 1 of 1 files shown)


opalmer:
@guidow, PTAL

opalmer:
@guidow, you get a chance to have a look at this?

guidow:
I'm going to have to keep an eye on whether or not automatic updates to free_ram still work, but other than that, this looks okay.

opalmer:
Thanks.  That's actually something I'm also concerned about and my initial guess is it might be an issue on hosts which don't have a lot of work running through them.  That said, this should clean things up and if it turns out to not be enough we can always schedule more frequent updates or include free_ram in other POST requests.

PR #221, opened 2015-02-23
Convert post_shutdown_to_master to inlineCallbacks

opalmer:
This change converts post_shutdown_to_master to inlineCallbacks.  In addition this change also fixes a couple of bugs regarding the shutdown timeout.

### DIFF
--- a/pyfarm/agent/service.py
+++ b/pyfarm/agent/service.py
@@ -526,7 +526,7 @@ class Agent(object):
 
             # We're hit the timeout
             if datetime.utcnow() > self.shutdown_timeout:
-                svclog.info("Shutdown timeout reached!")
+                svclog.error("Shutdown timeout reached!")
                 break
 
             for jobtype_id, jobtype in config["jobtypes"].copy().items():
--- a/pyfarm/agent/service.py
+++ b/pyfarm/agent/service.py
@@ -764,7 +764,6 @@ class Agent(object):
             self.shutdown_timeout = None
             return
 
-        self.shutdown_timeout = \
-            datetime.utcnow() + timedelta(
-                seconds=config["agent_shutdown_timeout"])
+        self.shutdown_timeout = timedelta(
+            seconds=config["agent_shutdown_timeout"]) + datetime.utcnow()
         svclog.debug("New shutdown_timeout is %s", self.shutdown_timeout)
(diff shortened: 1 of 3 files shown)


opalmer:
@guidow, PTAL.  I've finished adding tests for this.

guidow:
With these patches, the agent fails to shut down when requested. It will successfully post the changed state to the master, including retries if the master was not available at first, but will then fail to actually shut down.

opalmer:
Ok, that should be fixed now.  I had a feeling stop() may break but didn't get a chance to manually test it last night.  With this change stop() now just stops the agent instead of stopping the reactor too which is a little cleaner and it makes stop() more testable.  

I was planning to address convert `stop()` to inlineCallbacks in another branch but this seems like a good place to make the change since it's related.  But, I'm going to add tests for the new code before this merges (especially since stop()/sigint_handler() didn't have tests before)

opalmer:
@guidow, PTAL. There's tests for `sigint_handler()` and `stop()` now.  I made a few code changes to `stop()` but it should help in catching a few possible conditions we didn't before.

pyfarm/agent/service.py
```diff
@@ -472,151 +483,183 @@ def start(self, shutdown_events=True, http_server=True):
...
-                    reactor.stop()
+        svclog.info(
+            "Waiting on %s job types to terminate", len(config["jobtypes"]))
+
+        while True:
+            # No jobtypes remain
+            if not config["jobtypes"]:
+                break
+
+            # We're hit the timeout
```
guidow: *We have

pyfarm/agent/service.py
```diff
@@ -472,151 +483,183 @@ def start(self, shutdown_events=True, http_server=True):
...
-                    self.post_shutdown_to_master()
-                else:
-                    reactor.stop()
+        svclog.info(
+            "Waiting on %s job types to terminate", len(config["jobtypes"]))
+
+        while True:
+            # No jobtypes remain
+            if not config["jobtypes"]:
+                break
```
guidow: Can't you rewrite the while loop instead as
```while config["jobtypes"] and datetime.utcnow() <= self.shutdown_timeout```?

A ```while True``` in a place that's not actually supposed to be an endless loop seems kinda wrong to me...

guidow:
It works now.

pyfarm/agent/service.py
opalmer: Agreed, fixed.

opalmer:
Thanks for the review, since the suggested change is not functional and it didn't break the tests I'm going to go ahead and merge this.

Issue #222, opened 2015-02-24
Spurious newlines in tasklogs when running under Windows


guidow:
When running on Windows, the CSV-tasklogs will have one additional newline after every line.

Example:

    2015-02-24T14:00:57.250000,internal,1,Command: c:\Users\guido\stderr.bat

    2015-02-24T14:00:57.250000,internal,2,Arguments: ()

    2015-02-24T14:00:57.250000,internal,3,Work Dir: C:\

    2015-02-24T14:00:57.250000,internal,4,User/Group: None None

    2015-02-24T14:00:57.250000,internal,5,Environment:

    2015-02-24T14:00:57.250000,internal,6,{   }

    2015-02-24T14:00:57.394000,internal,7,"Started ProcessProtocol(uuid=44f4fa70-8af6-45ed-81d8-e79a48482a75, pid=640)"

    2015-02-24T14:00:57.492000,0,8,

    2015-02-24T14:00:57.492000,0,9,"C:\>echo ""Test""  1>&2 "

    2015-02-24T14:00:57.492000,1,10,"""Test"" "

    2015-02-24T14:00:57.498000,internal,11,"Process has terminated successfully, code 0"

guidow:
After some digging, it turns out the file that csv.writer writes to must be opened in binary mode on platforms where that makes a difference (i.e Windows).

See https://docs.python.org/2/library/csv.html#csv.writer

I'll make a patch tomorrow.

opalmer:
Thanks for looking into this, would not have expected that....Odd the docs don't really make a point of it but the associated PEP does seem to start to point it out:

https://www.python.org/dev/peps/pep-0305/#issues

opalmer:
Closing as the PR for this was merged.

PR #223, opened 2015-02-25
Fix spurious newlines in CSVs in Windows

guidow:
This also removes the indirection via in-memory file objects for writing
CSVs and the encoding feature for the csv output. This feature was never
actually used anywhere, and the resulting behaviour not relied upon.

Keeping the indirection makes it impossible to fix this bug.

### DIFF
--- a/pyfarm/agent/utility.py
+++ b/pyfarm/agent/utility.py
@@ -24,7 +24,6 @@ are copied over from the master (which we can't import here).
 
 import csv
 import codecs
-import cStringIO
 import os
 from decimal import Decimal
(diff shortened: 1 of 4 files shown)


guidow:
Addresses #222

opalmer:
LGTM.  The indirection was originally copied from Python's documentation but there's not much from the docs about why the indirection is necessary for unicode.  Like many modules in the standard library it may be to fully support a few edge cases in older Python versions.  I'd rather we fix this problem though so I'm going to go ahead and merge this.

PR #225, opened 2015-03-02
Remove File Utility

opalmer:
While looking around to figure out the next best code to transition to inlineCallbacks I found it would be easier to do a little cleanup first.  In this case I'm migrating calls to `os.remove` to a utility function which will file removal in a consistent manner while also offering some extra options.  The goal here is to provide a little cleanup while also not changing the intended behavior of the code being replaced.

This could also be used by job type implementors to handle file removal without having to write a bunch of wrapper code.


opalmer:
@guidow, PTAL

guidow:
Looks good, seems to work.

PR #226, opened 2015-03-02
Convert reannounce to inlineCallbacks

opalmer:
As the title says, this pull request converts `reannounce()` to use `@inlineCallbacks.`

### DIFF
--- a/pyfarm/agent/service.py
+++ b/pyfarm/agent/service.py
@@ -221,6 +221,8 @@ class Agent(object):
             return False
 
         contacted = config.master_contacted(update=False)
+        if contacted is None:
+            return True
         remaining = (datetime.utcnow() - contacted).total_seconds()
         return remaining > config["agent_master_reannounce"]
 
--- a/pyfarm/agent/service.py
+++ b/pyfarm/agent/service.py
@@ -86,13 +86,8 @@ class Agent(object):
         self.shutdown_timeout = None
         self.post_shutdown_lock = DeferredLock()
         self.stop_lock = DeferredLock()
+        self.reannouce_lock = DeferredLock()
         self.stopped = False
-
-        # reannounce_client is set when the agent id is
-        # first set. reannounce_client_instance ensures
-        # that once we start the announcement process we
-        # won't try another until we're finished
-        self.reannounce_client_request = None
 
         # Register a callback so self.shutdown_timeout is set when
         # "shutting_down" is set or modified.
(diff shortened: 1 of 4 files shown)


opalmer:
@guidow, PTAL tests are finished for this one.

pyfarm/agent/service.py
```diff
@@ -86,14 +86,9 @@ def __init__(self):
         self.shutdown_timeout = None
         self.post_shutdown_lock = DeferredLock()
         self.stop_lock = DeferredLock()
+        self.reannouce_lock = DeferredLock()
```
guidow: Should probably be ```self.reannounce_lock```. (The last "n" is missing.)

pyfarm/agent/service.py
```diff
@@ -222,83 +217,95 @@ def repeating_call(
...
+        return utility.total_seconds(
+            datetime.utcnow() - contacted) > config["agent_master_reannounce"]
 
+    @inlineCallbacks
     def reannounce(self, force=False):
         """
         Method which is used to periodically contact the master.  This
         method is generally called as part of a scheduled task.
         """
+        yield self.reannouce_lock.acquire()
```
guidow: Hm, can we maybe have a timeout on this lock? I wouldn't want a large number of waiting reannounce coroutines to accumulate over time if the master temporarily unavailable for one reason or another...

guidow:
Looks good and works in practice, but see inline comments.

pyfarm/agent/service.py
opalmer: Good point, even with the should_reannounce() check it could still be an issue.  I'l look into adding this, probably just need to override acquire() and raise a timeout exception of some kind.

opalmer:
Ok, added in a lock with a timeout.  I'll get some tests added for the new code here in the next day or so.

pyfarm/agent/service.py
```diff
@@ -222,83 +217,106 @@ def repeating_call(
...
         """
+        # Attempt to acquire the reannounce lock but fail after 70%
+        # of the total time between reannouncements elapses.  This should
+        # help prevent an accumulation of requests in the event the master
+        # is having issues.
+        try:
+            yield self.reannounce_lock.acquire(
+                config["agent_master_reannounce"] * .70
+            )
+        except utility.TimedDeferredLock:
```
guidow: Should probably be ```except utility.LockTimeoutError:```

opalmer:
@guidow, tests have been added.  This should be ready for a final review.

guidow:
The code looks good, but reannounces seem to be broken. I started the agent with --master-reannounce 3, and no reannounces are happening.

OTOH, I think they were broken even before this PR, so we might as well merge this.

opalmer:
Thanks, merging.  I'll try to take a quick stab to see if I can get that fixed in another branch.

PR #227, opened 2015-03-02
Improved Mocks In Tests

opalmer:
This is a minor update to some previous tests to use assert_called_with() when checking log statements.  It's less code in the tests and more directly verifies that we're logging what we expect.


opalmer:
@guidow, PTAL.  There's not any function changes here this time around.

opalmer:
I'm going to go ahead and merge this since it's not a functional change.

PR #228, opened 2015-03-05
Update the agent state to running when necessary

guidow:
Also, change it back to online when no assignment is being worked on
anymore.

As a side effect, this will cause the free_ram value to be updated
immediately after the last active assignment has stopped running,
which is probably the most interesting point in time for getting a
free_ram update anyway.


opalmer:
LGTM, merging.

PR #229, opened 2015-03-06
Agent state update bugfix

guidow:


setup.py
```diff
@@ -85,7 +85,7 @@ def get_package_data(parent, roots):
 
 setup(
     name="pyfarm.agent",
-    version="0.8.3",
+    version="0.8.3.1",
```
opalmer: 0.8.3 is not on PyPi or tagged yet.  Is there any reason we can't tag and release 0.8.3?  This would also keep the versioning scheme consistent across packages and between previous and future versions.

pyfarm/agent/http/api/assign.py
```diff
@@ -187,15 +187,15 @@ def assignment_started(_, assign_id):
 
         def assignment_stopped(_, assign_id):
             logger.debug("Assignment %s has stopped", assign_id)
+            if (len(config["current_assignments"]) <= 1 and
```
opalmer: Shouldn't this be `< 1` or `== 0`?  If `current_assignments` is 1 or higher I'd think the agent would be running.

guidow: No, because we do this immediately before removing the current assignment from config["current_assignments"], so it will always be at least 1.

setup.py
guidow: I did this version change because I was running git master on my agents, and I wanted to keep track of which agent already had this bugfix deployed. I can undo this agin if you object.

pyfarm/agent/http/api/assign.py
opalmer: Ah yes that makes sense thanks.

setup.py
opalmer: Yeah I think keeping the versioning scheme consistent with prior history and other packages would be best.

If it makes it easier for you, I'll be happy to release 0.8.3, merge this and then release 0.8.4.  That will make the development version 0.8.5 which might help with your tracking problem?

guidow: It's not a real problem anymore. I've just undone the version bump now.

opalmer: I'll go ahead and merge this then, I'll roll out 0.8.3 tomorrow unless you'd prefer I'd not.

opalmer: Done, this has been tagged and released.

PR #230, opened 2015-03-06
Fix shutdown when requested via API

guidow:


opalmer:
Looks like the tests are failing because the reactor is being stopped mid-test.  You'll probably need to update some of the API tests to mock out `reactor.stop()` then test to make sure it is in fact called.

opalmer:
Couple suggestions:

* You will probably need to mock out `reactor.stop()` in some of the api tests in order for this to work
* Use `stopping = config["agent"].sigint_handler()` instead of calling `stop()` directly.  That will handle stopping the agent, the reactor, return the proper exit code and remove the control file.

guidow:
> * Use `stopping = config["agent"].sigint_handler()` instead of calling `stop()` directly.  That will handle stopping the agent, the reactor, return the proper exit code and remove the control file.  If the behavior needs to vary slightly I'd rather we make a small update to `sigint_handler()` to handle it.

I need the control file to remain put in some of these cases, so this isn't really an option.

opalmer:
> I need the control file to remain put in some of these cases, so this isn't really an option.

Makes sense, mocking `reactor.mock()` would do fine then.  Looks like that's fixed so I'll go ahead and merge this.

PR #232, opened 2015-03-07
Remove Logging Mocks In remove_fie Tests

opalmer:
This change removes the mocks for logging in the tests for `remove_file()`.  They don't provide a great deal of benefit right and only complicate the test code.


opalmer:
Merging since this not a functional change or a change where we're changing the main purpose of the tests.

PR #233, opened 2015-03-07
remove_directory utility

opalmer:
This is a complimentary function to `utility.remove_file()`.  Adding this for use in tests mainly but also for job type implementors too.

### DIFF
--- a/pyfarm/agent/utility.py
+++ b/pyfarm/agent/utility.py
@@ -431,7 +431,7 @@ def remove_directory(
             keywords = dict(retry_on_exit=False, raise_=False)
             signature = (remove_directory, (path, ), keywords)
             if signature not in atexit._exithandlers:
-                atexit.register(remove_file, path, **keywords)
+                atexit.register(remove_directory, path, **keywords)
         else:
             logger.error(
                 "Failed to remove %s (%s)",
--- a/pyfarm/agent/utility.py
+++ b/pyfarm/agent/utility.py
@@ -370,8 +370,6 @@ def remove_file(
         os.remove(path)
     except (WindowsError, OSError, IOError) as error:
         if error.errno in ignored_errnos:
-            logger.debug(
-                "Failed to remove %s (%s)", path, errorcode[error.errno])
             return
 
         if retry_on_exit:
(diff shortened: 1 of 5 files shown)


opalmer:
@guidow, PTAL

pyfarm/agent/utility.py
```diff
@@ -392,3 +393,51 @@ def remove_file(
...
+    exit if requested.  This function standardizes the log output, ensures
+    it's only called once per path on exit and handles platform
+    specific exceptions (ie. ``WindowsError``).
+
+    :param bool retry_on_exit:
+        If True, retry removal of the file when Python exists.
+
+    :param bool raise_:
+        If True, raise an exceptions produced.  This will always be
+        False if :func:`remove_file` is being executed by :module:`atexit`
```
guidow: ```remove_file``` should probably be ```remove_directory```

pyfarm/agent/utility.py
```diff
@@ -392,3 +393,51 @@ def remove_file(
...
+
+    :param tuple ignored_errnos:
+        A tuple of ignored error numbers.  By default this function
+        only ignores ENOENT.
+    """
+    try:
+        shutil.rmtree(path)
+    except (WindowsError, OSError, IOError) as error:
+        if error.errno in ignored_errnos:
+            logger.debug(
```
guidow: I think logging that is overkill. We are already producing enough log output that getting important information out of it is a lot of work.

pyfarm/agent/utility.py
```diff
@@ -392,3 +393,51 @@ def remove_file(
...
+                "Failed to remove %s (%s)", path, errorcode[error.errno])
+            return
+
+        if retry_on_exit:
+            logger.warning(
+                "Failed to remove %s (%s), will retry on exit",
+                path, errorcode[error.errno])
+
+            # Make sure we're not added multiple times
```
guidow: Would that be a problem?

guidow:
Looks good to me, but see inline comments.

pyfarm/agent/utility.py
opalmer: Ah good point thanks, fixed.

pyfarm/agent/utility.py
opalmer: Fair point.  Removing it from here and from `remove_file` as well since the same comment would apply.

pyfarm/agent/utility.py
opalmer: Potentially over time yes because you could add the same path over and over by accident or in a loop.  The agent can run for quite a while so I'm trying to make sure we're not unnecessarily building up entries in a list that will persist until shutdown (which will also make the shutdown take longer).

PR #234, opened 2015-03-07
Fix testutil test location

opalmer:
There are some tests for testutil which are currently in the wrong location, this PR fixes that.


opalmer:
Merging, no code changes at all here.

PR #235, opened 2015-03-07
Tests for start_daemon_posix()

opalmer:
This change adds test coverage for `start_daemon_posix()`.


opalmer:
@guidow, PTAL

tests/test_agent/test_entrypoints_utility.py
```diff
@@ -14,3 +14,182 @@
...
+from random import randint
+
+from mock import Mock, patch
+
+from pyfarm.agent.testutil import TestCase, skipIf
+from pyfarm.agent.entrypoints import utility
+from pyfarm.agent.utility import remove_file
+
+
+class TestStartDamonPosix(TestCase):
```
guidow: *Daemon.

guidow:
I'm not a huge fan of this. You end up mocking out so many system functions that the function you end up testing bears little resemblance to what will actually run in production.

On top of that, the asserts in your tests really only test that the functions works "as coded", which is not really terribly useful. More like tautological.

If you're after increasing the coverage percentage on coveralls, then, sure, this might help, but I'm not convinced these tests are good for anything else, to be honest.

opalmer:
> the asserts in your tests really only test that the functions works "as coded"

That's pretty much what I was going for with this, increasing coverage was nice but not the main goal.  I  agree that in reality this would be hard to test which is what the TODO in ec8201513dfc0396257a2215a2c7df8fc38a7355 was for (which should be readded if we decided to keep these tests).

The tests here were mainly to ensure that if we updated this in the future we'd mostly be calling the same functions and in an order that seemed to make sense.  But, while writing this I had many of thoughts you're brining up now so it's probably not worth merging this.  I'd still like to find a way to test the behavior at some point though so maybe I'll try making another pass at this some other time.

PR #236, opened 2015-03-07
More test coverage for pyfarm.agent.utility

opalmer:
This PR does three things:
* Adds some missing tests for `pyfarm.agent.utility`
* Ensures that we also handle `WindowsError` when dealing with files
* Adds missing coverage exclusion for `logger.errror`


opalmer:
@guidow, PTAL

guidow:
Looks good.

PR #238, opened 2015-03-08
Improved coverage for system_data()

opalmer:
`system_data()` was missing a few lines of coverage, this PR fixes that.


opalmer:
Merging since `system_data()` already had tests (this is just adding missing coverage) and this PR is not changing the code itself.

PR #239, opened 2015-03-08
Improve HTTP Resources

opalmer:
This PR makes several related improvements to the core resource and related classes.  This change is also a prerequisite for moving to `inlineCallbacks` for a few of our HTTP apis. 

The main changes in this pull request are:

* Base `Resource` class now allows for tuple or deferred responses.  This simplifies many of the HTTP apis and will also allow us to return a deferred instance from a resource.
* Cleaned up and simplified the internals of `Resource.render()`
* Replaced the old template class with `EncodedStringTemplate` which:
  * Has a new `render()` method which encodes to the ascii character set.  This fixes a bug where method could return a unicode response which then causes other Twisted methods to fail (because they are expecting strings only.
  * Has a new `render()` method which no longer returns a Deferred.  The old implementation was not performing any non-blocking actions so returning a Deferred was unnecessary. 
* Replaced `FakeRequest` with Twisted's `DummyRequest` which is closer to how a real request works.
* Removed the old `RewriteRequest` as it was masking possible bugs in the code.


opalmer:
@guidow, PTAL.

opalmer:
@guidow, it would probably be best to look and/or test this one first over #226 if you can only direct take on one at a time.  One of the next places I'm looking to improve and add coverage for are the HTTP apis so this change would need to land first.

pyfarm/agent/http/core/resource.py
```diff
@@ -87,40 +79,54 @@ def template(self):
...
+        return tuple(methods)
+
+    def response_types(self, request, default=None):
+        """
+        Returns an instance of :class:`frozenset` which contains the type
+        or types of responses we can send.  We look at the ``Accept`` and
+        ``Content-Type`` headers to determine this information.  If values
+        are present in the ``Accept`` header these will be returned instead
+        of values in ``Content-Type``.  If neither of these headers are present
+        we'll return what was provided to ``default`` instead.
```
guidow: The ```Content-Type``` header is NOT supposed to be used in this way.

Quote from the RFC:

     If no Accept header field is present, then it is assumed that the client accepts all media types.

pyfarm/agent/http/core/resource.py
```diff
@@ -129,90 +135,152 @@ def error(self, request, code, message):
...
     def render(self, request):
-        # make sure that the requested content type is supported
-        content_type = self.content_types(request, default=["text/html",
-                                                            "application/json"])
-        if not self.CONTENT_TYPES & content_type:
+        # Check to ensure that the content type being requested
+        content_types = self.response_types(
+            request, default=["text/html", "application/json"])
+
+        if not self.SUPPORTED_CONTENT_TYPES & content_types:
```
guidow: The ```Accept``` header can contain wildcards, and the master does use that when contacting the agent. This doesn't handle that case.

pyfarm/agent/http/core/template.py
```diff
@@ -55,32 +55,25 @@ def dump_bytecode(self, bucket):  # pragma: no cover
...
-            # get the results then convert to a string, Twisted can't handle
-            # unicode from here
-            deferred.callback(
-                str(super(DeferredTemplate, self).render(*args, **kwargs)))
-        except Exception as e:  # pragma: no cover
-            deferred.errback(e)
-
-        return deferred
+        result = super(EncodedStringTemplate, self).render(*args, **kwargs)
+        return result.encode("ascii", "replace")
```
guidow: Could you explain this part more? Why are we forcing everything to ascii here?

pyfarm/agent/http/core/resource.py
```diff
@@ -129,90 +135,152 @@ def error(self, request, code, message):
...
+        assert isinstance(deferred, Deferred)
+        response = yield deferred
+        self.render_tuple(request, response)
+
     def render(self, request):
-        # make sure that the requested content type is supported
-        content_type = self.content_types(request, default=["text/html",
-                                                            "application/json"])
-        if not self.CONTENT_TYPES & content_type:
+        # Check to ensure that the content type being requested
```
guidow: This sentence is incomplete.

pyfarm/agent/utility.py
```diff
@@ -196,7 +196,11 @@ def dumps(*args, **kwargs):
     else:
         obj = dict(*args, **kwargs)
 
-    return _dumps(obj, default=default_json_encoder, indent=indent)
+    dump = _dumps(obj, default=default_json_encoder, indent=indent)
+
+    # Many parts of Twisted will not accept the typical output
+    # from _dumps, which is unicode, so we encode into ascii here.
+    return dump.encode("ascii", "replace")
```
guidow: That doesn't seem right to me. We should be able to send data in any encoding we want, as long as we present it as a stream of bytes. Limiting ourselves to ascii because of a limitation in Twisted seems seriously wrong to me.

guidow:
This doesn't work like this. When I run the agent with this branch, and the master sends an assignment to it, I get this error:

    2015-03-10 16:07:10 ERROR    - pf.agent.http.resource - frozenset(['*/*']) is not a support content type for this url
    2015-03-10 16:07:10 DEBUG    - HTTPChannel,0,127.0.0.1 - "127.0.0.1" - - [10/Mar/2015:15:07:09 +0000] "POST /api/v1/assign HTTP/1.1" 415 67 "-" "PyFarm/1.0 (master)"

pyfarm/agent/utility.py
opalmer: Agreed and I should have looked closer.  This turns out to be an artifact of the `DummyRequest` class which is forcing a check on the type of data being written.  I've fixed this in e13c5fc.

pyfarm/agent/http/core/template.py
opalmer: Rolled back in e13c5fc757ee5a4d3533c804d4033f99f421519a.  This was in response to a behavior that the `DummyRequest` class in `write()`.

opalmer:
Ok, still working on updating the tests for these changes but I think that code wise this this should be better.  I think the `Content-Type`/`Accept` headers are better handled now and we should have the control we need to properly accept cases like `""` for Content-Type or '*/*` for accept.

pyfarm/agent/http/core/resource.py
```diff
@@ -129,90 +175,169 @@ def error(self, request, code, message):
...
-                                                            "application/json"])
-        if not self.CONTENT_TYPES & content_type:
+        assert self.SCHEMAS is not NotImplemented
+        assert self.ALLOWED_ACCEPT is not NotImplemented
+        assert self.ALLOWED_CONTENT_TYPE is not NotImplemented
+
+        # Ensure we can handle the content of the request
+        content_types = \
+            self.get_content_type(request) & self.ALLOWED_CONTENT_TYPE
+        if not content_types:
```
guidow: This check must not be done if the incoming request does not have a body, which is the norm for GET, DELETE, and HEADER requests.

guidow:
Still doesn't work. This time, all requests that do not contain a body result in an error.

opalmer:
Updated (still have tests to fix).  Odd though, the `""` value in `DEFAULT_CONTENT_TYPE` should have prevented that from happening because `get_content_type()` would have returned `DEFAULT_CONTENT_TYPE` if the header was not present or empty.

Regardless, checking to see if there's any content at all would be better to do first anyway...

guidow:
It's still broken:

Start the agent, go to http://127.0.0.1:50000. You get an error message.

Here's the captured traffic between the browser and the agent:

    GET / HTTP/1.1
    Host: 127.0.0.1:50000
    User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:36.0) Gecko/20100101 Firefox/36.0
    Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
    Accept-Language: en-US,en;q=0.5
    Accept-Encoding: gzip, deflate
    Connection: keep-alive
    Cache-Control: max-age=0
    
    HTTP/1.1 415 Unsupported Media Type
    Transfer-Encoding: chunked
    Date: Thu, 12 Mar 2015 09:50:59 GMT
    Content-Type: text/html
    Server: TwistedWeb/15.0.0
    
    43
    {
      "error": "Can only handle text/html or application/json here"
    }
    0

Note also how the Content-Type for the response is given as ```text/html``` even though it's actually JSON.

opalmer:
I'm going to take a deeper look at this, and requests to the API, over the weekend.  This week was busy so I didn't get the time I needed to invest in fully testing this after work, hopefully I'll have this working for on Monday...

opalmer:
Ok, I think I've fixed that error now.  There's still a few more to work through though.

opalmer:
Ok, I think I've fixed that error now.  Looks like the tests are passing too as well so this is probably worth another look in terms of code.  I'm going to do some more manual testing tomorrow of my own as well...

guidow:
The bugs have been ironed out now, it works again.

opalmer:
Cool, you want to continue testing this for a bit or should I go ahead and merge?

guidow:
If you think this is ready, I think we can merge this now.

opalmer:
Merging, did some additional testing this weekend so I think we're good for this one.

PR #240, opened 2015-03-08
Fix assertion in service test

opalmer:
There's no such thing as `assert_never_called()`, use `self.assertFalse(<thing>.called)` instead.


opalmer:
Merging, this is a test fix only.

PR #241, opened 2015-03-10
jobtype.core.internals basic test improvements and bugfixes

opalmer:
This PR improves on the tests for `jobtype.core.internals` and will provide a couple bugfixes as well.

### DIFF
--- a/pyfarm/jobtypes/core/internals.py
+++ b/pyfarm/jobtypes/core/internals.py
@@ -806,8 +806,11 @@ class System(object):
         :raises InsufficientSpaceError:
             Raised if enough space cannot be claimed.
         """
-        assert isinstance(tempdir, STRING_TYPES)
-        assert isinstance(space, INTEGER_TYPES)
+        assert isinstance(tempdir, STRING_TYPES), "Expected string for tempdir"
+        try:
+            space = int(space)
+        except (ValueError, TypeError):
+            raise TypeError("Could not convert %r to an integer" % space)
 
         try:
             os.makedirs(tempdir)
--- a/pyfarm/jobtypes/core/internals.py
+++ b/pyfarm/jobtypes/core/internals.py
@@ -712,9 +712,6 @@ class System(object):
         """
         Internal function which handles both user name and group conversion.
         """
-        if not isinstance(value, STRING_TYPES):
-            raise TypeError("Expected string for `value`")
-
         # This platform does not implement the module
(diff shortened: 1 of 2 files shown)


opalmer:
@guidow, PTAL.  These are some pretty standalone fixes and I figured this would be a good place for a review/merge.  There's still a fair amount of code to go through coverage wise in `internals.py` but I'm going to do that in another branch since I'm also going to be updating some of the code with `inlineCallbacks`

guidow:
The code looks good, and it works in practice.

guidow:
Sorry, but I have to withdraw the OK again – I just deployed this patch set to one of the nodes in our test farm, and it would immediately fail running assignments with an AssertionError on starting the jobtype.

I don't yet know which assert has failed there and why. Will have to investigate further tomorrow.

opalmer:
Thanks, take your time.  If I had to guess it's probably either one of the assertions I added or it's related to the property change somehow...

pyfarm/jobtypes/core/internals.py
```diff
@@ -774,20 +787,31 @@ def _remove_tempdirs(self):
...
+        .. warning::
+
+            This will delete files in tempdir to reclaim storage space.
 
-        Will raise InsufficientSpaceError if enough space cannot be claimed.
+        :raises InsufficientSpaceError:
+            Raised if enough space cannot be claimed.
         """
+        assert isinstance(tempdir, STRING_TYPES)
+        assert isinstance(space, INTEGER_TYPES)
```
guidow: Turns out, this was the problematic assert. The jobtype specified the required space as ```5e9```, which used to work just fine without this assert, but now fails because it's technically a float, not an int.

I suggest that, instead of the assert, you just convert the value to int and let this raise the appropriate exception if that fails.

Also, it would be nice if the asserts had an error message that tells us where the AssertionError came from when we catch it. Would make debugging a lot easier.

opalmer: Fixed, PTAL

guidow:
The agent now does its work again with the latest commit.

opalmer:
Updated, thanks.  I wrapped the exception because I wanted it to consistently be a `TypeError` and I thought we'd benefit from a better error message than what Python would normally provide.  I'll go ahead and merge this tomorrow unless you have more comments you'd like to make.

opalmer:
My bad, fixed.  Merging.

PR #242, opened 2015-03-12
Bugfix: Set agent state on failed assignment, too

guidow:
Previously, the agent would erroneously leave its state on running if an
assignment stop with a failure.


opalmer:
LGTM, merging.

PR #243, opened 2015-03-12
Upload logs after the assignment has finished

guidow:
This avoids a situation where the log has not been closed yet and might
still be written to while being uploaded.

As a side effect, _upload_logfile() now no longer takes a log
identifier, since an assignment is only supposed to have one log anyway.


opalmer:
LGTM, merging.

PR #244, opened 2015-03-17
Include farm_name in agent's data if specified

guidow:


opalmer:
Code looks ok, config needs to be updated though to include the new value.

guidow:
Hm, how do I include this so it defaults to ```None```?

opalmer:
Update `pyfarm/agent/etc/agent.yml` with something like this would do it:

```yaml
# Including a comment above the value would be nice too so an administrator
# does not have to go looking elsewhere.
farm_name: null
```

opalmer:
Looks good, thanks for the update.  Merging.

PR #245, opened 2015-03-20
Catch exceptions from JobType.is_successful()

guidow:
JobType.is_successful() is user-overridable, so we have to assume it
might raise an exception.

The previous code would fail to call the stopped deferred is that
happened, resulting in the agent hanging.

### DIFF
--- a/pyfarm/jobtypes/core/internals.py
+++ b/pyfarm/jobtypes/core/internals.py
@@ -456,7 +456,7 @@ class Process(object):
         process_data = self.processes.pop(protocol.uuid)
 
         try:
-            successful = self.is_successful(reason)
+            successful = self.is_successful(protocol, reason)
         except Exception as e:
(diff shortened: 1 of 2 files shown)


pyfarm/jobtypes/core/internals.py
```diff
@@ -455,7 +455,15 @@ def _process_stopped(self, protocol, reason):
         logger.info("%r stopped (code: %r)", protocol, reason.value.exitCode)
         process_data = self.processes.pop(protocol.uuid)
 
-        if self.is_successful(reason):
+        try:
+            successful = self.is_successful(protocol, reason)
+        except Exception as e:
+            logger.error("Exception caught from is_successful: %s: \"%s\". "
```
opalmer: You could construct the error message once and use it in both places.  I'd also use `%r` instead of manually escaping quotes.

opalmer:
LGTM, merging.

PR #247, opened 2015-03-24
Encode only unicode objects to UTF-8

### LINKED ISSUE #246: Incomplete tasklogs
Sometimes the agent will produce tasklogs which are missing several lines at the end. This happens rarely, but it does happen from time to time. I have not yet found a way to reliably reproduce this, but this jobtype seems to produce it more often than others (on Windows):

    from twisted.python.procutils import which

    from pyfarm.jobtypes.core.jobtype import JobType, CommandData

    class SleepFiveMinutes(JobType):
        def get_command_data(self):
            ping = which("ping")

            return CommandData(ping[0], "-n", "301", "127.0.0.1")

        def is_successful(self, protocol, reason):
            self._log("In is_successful, return code: %s" % reason.value.exitCode)
            if reason.value.exitCode == 0:
                return True
            else:
                return False

guidow:
"str" objects will be left as they are.

Fixes #246

### DIFF
--- a/pyfarm/agent/utility.py
+++ b/pyfarm/agent/utility.py
@@ -250,7 +250,8 @@ class UnicodeCSVWriter(object):
         self.writer = csv.writer(f, dialect=dialect, **kwds)
 
     def writerow(self, row):
-        self.writer.writerow([s.encode("utf-8") for s in row])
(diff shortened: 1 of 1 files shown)


opalmer:
Looks good, merging.

PR #248, opened 2015-03-24
Rework how LoggerPool.flush() writes to disk

guidow:
Simplifies LoggerPool.flush() a bit.  Now only one thread will ever
write messages to the log file at a time, until it is finished.
Before, if there were several threads writing (which can happen if a lot
of output is produced in a short time), the lock would constantly pass
from one thread to the next, and as long as messages keep coming in, no
thread will ever actually finish.

Also, now the log file is unconditionally flushed every time a thread
ends.  This is better for jobtypes that produce little output in case an
admin wants to find out how far a running assignment has gotten (or why
it is stuck).

### DIFF
--- a/pyfarm/jobtypes/core/log.py
+++ b/pyfarm/jobtypes/core/log.py
@@ -57,7 +57,6 @@ class CSVLog(object):
         self.file = fileobj
         self.messages = deque()
         self.lines = 0
-        self.written = 0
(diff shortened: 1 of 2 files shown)


opalmer:
Nice, this will also keep resource usage a little more consistent and not be as heavy on the lock.  Looks good, merging.

PR #250, opened 2015-03-26
Improve error message in setup.py for Windows

opalmer:
This PR is aimed at improving the information we're displaying to someone who is trying to install the agent on Windows.  This was opened in response to 
https://groups.google.com/a/pyfarm.net/forum/#!msg/discuss/PH6YkHV860U/4aYcXpULXP0J

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,32 +38,13 @@ if "READTHEDOCS" in os.environ:
 # installed.  It's a requirement of Twisted, mainly because of spawnProcess,
 # however other parts of Twisted use it as well.  Twisted's setup.py itself
 # does not declare this dependency, likely because of the difficulties
-# associated with installing the package.
+# associated with installing the package.  Eventually Twisted will move
+# away from this requirement once https://twistedmatrix.com/trac/ticket/7477
+# is closed.  In the mean time however we'll use pypiwin32 which is built
+# by some of Twisted's maintainers:
+#   http://sourceforge.net/p/pywin32/feature-requests/110/
 if sys.platform.startswith("win"):
-    try:
-        import win32api
-    except ImportError:
-        raise ImportError(
-            "\r\nOn Windows, you must manually install pywin32 before running "
-            "pyfarm.agents's setup.py.  This is not a package that can "
-            "reliably be installed from PyPi so there's three options here:\r\n"
-            "  * Download and run one of the official installers:\r\n "
-            "    http://sourceforge.net/projects/pywin32/files/pywin32/\r\n"
-            "  * Install from a prebuilt binary wheel:\r\n"
-            "    http://www.lfd.uci.edu/~gohlke/pythonlibs/#pywin32\r\n"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -35,20 +35,34 @@
     install_requires += ["sphinxcontrib-httpdomain", "sphinx"]
 
 # Windows is a little special because we have to have pywin32
-# installed.  pyfarm.core uses it and certain components of
-# other libraries use it too, such as twisted, so we check for
-# it here.  Unfortunately, we can't use PyPi for this.
+# installed.  It's a requirement of Twisted mainly because of spawnProcess
```
guidow: I think this is missing a comma.

guidow:
Hm, I don't know how I feel about telling users to get a precompiled binary package from yet another third party...

opalmer:
To be honest, I don't like doing it either especially since it's an extra step for Windows.  Perhaps, with permission from the person who built the wheel file, we host the wheel ourselves and install it along with the agent?

opalmer:
Any more comments on this one?  I know it's not ideal, but I'm not sure we have a great deal of options right now.  I'm looking into something like ctypes or maybe contributing [cffi bindings](https://twistedmatrix.com/trac/ticket/7477) to Twisted so hopefully at some point we can remove these suggestions.

guidow:
I still don't think we should make getting the package from some third-party packager the default "if-unsure" recommendation.

My problems with the installer might just have been because I did not have admin rights when running it. (If that's it, we may need to file a bug with the pywin32 project, though. That error message is seriously unhelpful.) With the wheels from Christian Gohlke, you still have to do a manual step after the install that requires admin rights.

opalmer:
So after looking around some more, I think the last few commits offer a better solution.  You'll still have to install the wheel package but the only command you'd have to run is `pip install wheel`.  After that, the setup.py can pull in the [pypiwin32 package](https://pypi.python.org/pypi/pypiwin32/) which a couple of the core Twisted developers seem to be the maintainers of [as a result of this feature request](http://sourceforge.net/p/pywin32/feature-requests/110/).

Since we don't need most of pywin32, and neither does Twisted, I might take a look at an open bug they have to remove the dependency.  In the mean time however this is probably the best solution we have for Windows.

opalmer:
So after looking around some more, I think the last few commits offer a better solution.  You'll still have to install the wheel package but the only command you'd have to run is `pip install wheel`.  After that, the setup.py can pull in the [pypiwin32 package](https://pypi.python.org/pypi/pypiwin32/) which a couple of the core Twisted developers seem to be the maintainers of [as a result of this feature request](http://sourceforge.net/p/pywin32/feature-requests/110/).

Since we don't need most of pywin32, and neither does Twisted, I might take a look at an open bug they have to remove the dependency.  In the mean time however this is probably the best solution we have for Windows.

guidow:
Sounds good to me.

opalmer:
Great, merging.

PR #252, opened 2015-03-28
TypeChecks Cleanup/Improvements

opalmer:
These is a non-functional change that makes some small improvements to the `TypeChecks` class.  The main change is switching to `@staticmethod` since we don't need access to either the class or the instance.  This change also removes some tests which are no longer valid or are already covered by other tests.


opalmer:
Merging since this won't produce any functional change.

PR #253, opened 2015-03-30
Remove unused stop_reactor keyword

opalmer:


opalmer:
Merging, there are not any functional changes here and grepping around for stop_reactor yielded zero results.

PR #254, opened 2015-03-30
Improve test_http_api_state.TestStatus

opalmer:
`TestStatus` has been somewhat unreliable and has caused failures in the past when we get unexpected results.  This change mocks out `memory.free_ram()` and `time.time()` so they return constant values.   In this particular test we don't care that those values are accurate we just care that they meet our expected results.  This change also properly cleans up the config in between test runs.


opalmer:
Merging, this is only making a test less-flaky.

PR #255, opened 2015-03-30
Change Config Logging to Debug

opalmer:
This change changes all logging calls in the config object to debug to cut down on logging output.  Full comment from 7709711506affb0e2f8fc02155168d2efa79d3c0:

```
The agent's configuration logger is probably
one of the most verbose loggers.  Its output
is important but probably only in cases where
there's a problem being debugged.  Even then the
configuration is usually not going to be the cause
of the issue so keeping the configuration logger
in debug only mode probably makes more sense.
```


opalmer:
@guidow, PTAL.

guidow:
I fully support this. I logging output is way too much atm.

PR #256, opened 2015-03-30
Configuration Based Logger Level

opalmer:
This change adds the `agent_global_logger_level` configuration variable.  The main purpose of this variable is to provide some control over the global logging level.  

The only side effect that I know of with this change is we'll miss a few debug level messages from just prior to the config loading because the default level is now INFO in `twistd.py`.  This is probably fixable with a couple of work arounds but I don't think we're going to be missing much because the config module is almost always one of the first modules to load.


opalmer:
@guidow, PTAL

guidow:
Looks good.

PR #257, opened 2015-03-30
Don't reset finished tasks in JobType.stop()

guidow:


pyfarm/jobtypes/core/jobtype.py
```diff
@@ -739,7 +739,8 @@ def stop(self, assignment_failed=False, error=None, signal="KILL"):
                         "state to failed", task["id"])
         else:
             for task in self.assignment["tasks"]:
-                if task["id"] not in self.failed_tasks:
+                if (task["id"] not in self.failed_tasks and
```
opalmer: A little cleaner IMHO:
```python
if task["id"] not in (self.failed_tasks, self.finished_tasks):
```

opalmer:
Looks good, thanks.  Merging.

PR #258, opened 2015-04-01
Include a traceback when a jobtype cannot start

guidow:


pyfarm/agent/http/api/assign.py
```diff
@@ -25,6 +25,7 @@
         ACCEPTED, BAD_REQUEST, CONFLICT, SERVICE_UNAVAILABLE, OK)
 
 from functools import partial
+import traceback
```
opalmer: For consistency with other files, this import should go above functools:

```python
import traceback
from functools import partial
```

opalmer:
Looks good, thanks.  Merging.

PR #260, opened 2015-04-02
Improve testutil setUp

opalmer:
The main points behind this PR are included in commit 7da7052b2075df19d8646c8152508249b5330dd4's comment.

* removing the DNS hostname check, we should just
  let the request fail instead
* Move the url check into setUp() so it's not
  performed at startup
* If the configuration does not define a url, don't
  run the tests that require the url
* Set a timeout on the socket so we don't sit there forever


opalmer:
@guidow, PTAL.  I was having a couple of network issues, which turned out to be a firewall problem in the end, and came up with these improvements.  These seem to handle times a little better and allow for certain remote tests to be disabled if you override them (which could be useful when no internet connection exists).

guidow:
Looks good.

PR #251, opened 2015-03-28
[WIP] Update JobType Cache and Assignment Endpoint to use inlineCallbacks

opalmer:
This PR reworks the internal cache for jobtypes.  The main goals of this PR are to:

* Convert to `inlineCallbacks`
* Improve caching behavior, especially between cached and uncached modes.
* Fix a few subtle bugs in the jobtype cache.
* Update the assignment endpoint to use the new  `inlineCallbacks` to match the new cache implementation

PR #261, opened 2015-04-02
Don't let improper log formatting raise exceptions

opalmer:
Found this bug while working on #251, figured it should be split into its own review.


opalmer:
@guidow, PTAL

opalmer:
I'm going to go ahead and merge this since it unbreaks a couple of tests I'm working on and does not functionally change anything.

Issue #249, opened 2015-03-26
The first treq request after post_shutdown_to_master() fails


guidow:
I have been trying to change the Agent.stop() method so that it will set itself to offline first and then stop running assignments (because otherwise the master would assign new tasks to the agent in the short time inbetween).

When I did this, though, the state update for the first task in the assignment failed with this error:

    2015-03-26 12:20:29 ERROR    - pf.jobtypes.core - Error while posting state update for task, [Failure instance: Traceback (failure with no frames): <class 'twisted.web._newclient.RequestTransmissionFailed'>: [<twisted.python.failure.Failure <class 'twisted.internet.error.ConnectionDone'>>]                                                                            
    ], retrying                                                                                                                                                                            
    2015-03-26 12:20:29 ERROR    - root            - Unexpected exception from twisted.web.client.FileBodyProducer.stopProducing
    Traceback (most recent call last):                                                                                                                                                     
      File "/var/tmp/virtualenv/pyfarm-agent-python2/lib/python2.7/site-packages/Twisted-15.0.0-py2.7-linux-x86_64.egg/twisted/internet/endpoints.py", line 128, in connectionLost         
        return self._wrappedProtocol.connectionLost(reason)                                                                                                                                
      File "/var/tmp/virtualenv/pyfarm-agent-python2/lib/python2.7/site-packages/Twisted-15.0.0-py2.7-linux-x86_64.egg/twisted/web/_newclient.py", line 911, in dispatcher                 
        return func(*args, **kwargs)                                                                                                                                                       
      File "/var/tmp/virtualenv/pyfarm-agent-python2/lib/python2.7/site-packages/Twisted-15.0.0-py2.7-linux-x86_64.egg/twisted/web/_newclient.py", line 1564, in _connectionLost_TRANSMITTING                                                                                                                                                                                     
        self._currentRequest.stopWriting()                                                                                                                                                 
      File "/var/tmp/virtualenv/pyfarm-agent-python2/lib/python2.7/site-packages/Twisted-15.0.0-py2.7-linux-x86_64.egg/twisted/web/_newclient.py", line 812, in stopWriting                
        _callAppFunction(self.bodyProducer.stopProducing)                                                                                                                                  
    --- <exception caught here> ---                                                                                                                                                        
[...]

PR #263, opened 2015-04-07
Register logfile before starting assignment

guidow:
If we set the state of tasks not just on the task itself, but also on the current tasklog, we must make sure the tasklog is registered before we do that, otherwise that will fail with a 404.

In extreme cases, where assignments finish very fast, even uploading the final tasklog file at the end may fail because of this. (This can happen in practice if a jobtype class decides that no process needs to be started.)

Unfortunately, this PR still has a problem: When setting our agent's state, the request fails with this message:

    2015-04-07 17:36:57 INFO     - pf.jobtypes.core - Spawning Command(command='/usr/bin/blender', arguments=('-b', '/home/guido/stift_aniex43_finish.blend', '-s', '196.0', '-e', '200.0', '-o', '/tmp/stift_####.exr', '-a'), cwd='/home/guido/git/pyfarm', user=None, group=None, env={})
    2015-04-07 17:36:57 ERROR    - pf.agent.http.client - POST http://127.0.0.1:5000/api/v1/agents/a113b5a8-9822-413b-b1fd-fce3014956cf has failed (uid: 2d57ff05bd4f):
    Traceback (most recent call last):
    Failure: twisted.web._newclient.ResponseNeverReceived: [<twisted.python.failure.Failure <class 'twisted.internet.error.ConnectionDone'>>]

    2015-04-07 17:36:57 ERROR    - pf.agent.service - Failed to announce self to the master: [<twisted.python.failure.Failure <class 'twisted.internet.error.ConnectionDone'>>].  Will retry in 1.96419640754 seconds.

It works on retry, but I have no idea where this problem comes from.

### DIFF
--- a/pyfarm/jobtypes/core/internals.py
+++ b/pyfarm/jobtypes/core/internals.py
@@ -292,46 +292,6 @@ class Process(object):
     def __init__(self):
         self.start_called = False
         self.stop_called = False
-        self._stopped_deferred = None
-        self._started_deferred = None
-
-    @property
-    def stopped_deferred(self):
-        if not self.start_called:
-            raise RuntimeError("Not yet started")
-        return self._stopped_deferred
-
-    @property
-    def started_deferred(self):
-        if not self.start_called:
-            raise RuntimeError("Not yet started")
-        return self._started_deferred
-
-    @stopped_deferred.setter
-    def stopped_deferred(self, value):
-        if not self.start_called:
-            raise RuntimeError("Not yet started")
-
-        if self._stopped_deferred:
-            raise ValueError("Deferred already set")
-
-        if not isinstance(value, Deferred):
-            raise TypeError("Expected a Deferred instace")
-
-        self._stopped_deferred = value
-
-    @started_deferred.setter
-    def started_deferred(self, value):
-        if not self.start_called:
-            raise RuntimeError("Not yet started")
-
-        if self._started_deferred:
-            raise ValueError("Deferred already set")
-
-        if not isinstance(value, Deferred):
-            raise TypeError("Expected a Deferred instace")
-
-        self._started_deferred = value
 
     def _before_spawn_process(self, command, protocol):
         logger.debug(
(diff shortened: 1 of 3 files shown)


pyfarm/jobtypes/core/internals.py
```diff
@@ -292,46 +292,6 @@ class Process(object):
     def __init__(self):
         self.start_called = False
         self.stop_called = False
-        self._stopped_deferred = None
-        self._started_deferred = None
-
-    @property
-    def stopped_deferred(self):
```
opalmer: Could you go over the specifics as for why we're getting rid of these properties?  These are setup to try and ensure it's harder to missorder the setup of start_deferred/stop_deferred.

guidow: Since Process._start() will now register the tasklog before actually starting the process, it will have to return these two properties before the process has actually started. So, yeah, in this case we will have to access both of these before self.start_called is set to True.

guidow:
As for the problems described in the opening comments, I think I've gotten to the bottom of those, as well as, probably, the issues in #249. This seems to be a bug in Twisted-web (which treq uses as its backend), where if you do two consecutive, non-overlapping http requests with data against a server that doesn't support keep-alive, but doesn't make that explicit by setting a ```Connection: close``` header, the second request will fail because twisted tries to reuse the connection not realizing the server has closed it already...

See: https://twistedmatrix.com/trac/ticket/7843
and: http://twistedmatrix.com/pipermail/twisted-web/2015-April/005139.html

Since, in production, farms will probably use a real webserver instead of flask's built-in standalone mode, this doesn't seem like a major issue. Besides, as long as we're restarting the request when this happens, we're fine anyway.

opalmer:
Interestingly the problem you're describing here is something I was able to replicate at work.  I basically came to the same conclusions you did by accident when I did `data = treq.json_content(response)` when I meant to do `data = yield treq.json_content(response)` (which caused overlapping requests in my case).  I wanted to reply to your thread on twisted-web but was not subscribed at the time unfortunately.  Regardless, nice work and thanks for taking a deep look at that.

Anyway I'm going to go ahed and merge this, I think I have all my questions answers and I don't see anything code wise to fix here right now.

PR #264, opened 2015-04-09
Log file and directory deletions at level DEBUG

guidow:
Deleting files and directories is a normal part of operations.  There is
no reason to log that at log level info


opalmer:
Agreed, merging.  If there are particular directories that someone believe should be above debug, they can always log them themselves.

PR #265, opened 2015-04-09
Retry reannounce request if connection was dropped

guidow:
Workaround for https://twistedmatrix.com/trac/ticket/7843

### DIFF
--- a/pyfarm/agent/etc/agent.yml
+++ b/pyfarm/agent/etc/agent.yml
@@ -130,6 +130,13 @@ agent_http_retry_delay_factor: 5
 # be False.
 agent_http_persistent_connections: True
 
+# When using persistent connections, twisted will sometimes run requests over
+# connections that have already been closed by the server without realizing it.
+# When this happens, we catch the resulting failure and retry the request.
+# We do this up to $agent_broken_connection_max_retry times, assuming that if
+# it still fails at that point, there is probably something else wrong.
+broken_connection_max_retry: 20
+
 # If True then html templates will be reloaded with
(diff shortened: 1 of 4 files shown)


opalmer:
Code looks good, but could we get a configuration variable for the number of retries for these errors?  We're already hard coded the value twice and will probably need it again elsewhere.

pyfarm/agent/etc/agent.yml
```diff
@@ -130,6 +130,13 @@ agent_http_retry_delay_factor: 5
 # be False.
 agent_http_persistent_connections: True
 
+# When using persistent connections, twisted will sometimes run requests over
+# connections that have already been closed by the server without realizing it.
+# When this happens, we catch the resulting failure and retry the request.
+# We do this up to $agent_broken_connection_max_retry times, assuming that if
```
opalmer: `$agent_broken_connection_max_retry` -> `broken_connection_max_retry`?

I'm ok with either one, the doc string just needs to match the key.  Normally I'd suggest `agent_broken_connection_max_retry` for consistency with other vars. but since this is more for internal use I don't think the prefix is required.

opalmer:
Looks good except for the one comment above, thanks for the update.  Once that's fixed feel free to merge this.

pyfarm/agent/etc/agent.yml
guidow: I started with ```agent_broken_connection_max_retry```, but then decided that was too long and went with ```broken_connection_max_retry```. It seems I forgot to change the doc string as well.

PR #266, opened 2015-04-09
Move posting shutdown forward in stop sequence

guidow:
By posting the state update before stopping running assignments, we
avoid having the master assign us new work in the narrow window
in-between, and then failing it when polling us the next time.


opalmer:
Sounds logical to me, merging.

PR #267, opened 2015-04-13
Fix deferred handling in register logfile method

guidow:
This fixes two bugs:

- The error callback was called with the wrong number of arguments.

- On retrying, this method would generate a new deferred and forget
  about the old one.


opalmer:
Thanks, good catch.

PR #268, opened 2015-04-13
Fix handling errors from jobtypes when starting

guidow:
Merging #263 ended up breaking handling exceptions the jobtype's start()
method.  This should fix it again.


opalmer:
Looks good, merging.

PR #270, opened 2015-04-14
Minor grammar fix in setup.py

guidow:


opalmer:
Oops my bad, sorry about that.  Merging.

PR #273, opened 2015-04-24
Remove assignment only after all state updates

guidow:
If we don't do this, the agent will sometimes reannounce itself with an
empty tasks list before all the tasks have been set to done/failed,
leading the master to believe the assignment has failed.


guidow:
Any objections left to merging this?

opalmer:
Nope none...I must have missed merging it myself or replying saying to go ahead.  I always add the tags/milestone once I've had a look so I missed a step.  Sorry about that, merge when ready.

PR #274, opened 2015-04-28
Refuse to set tasks to failed if shutting down

guidow:
If we are shutting down, the task most likely has not actually failed,
and the child process(es) was/were killed because of the shutdown.  This
should not count towards the failure count of the current tasks.


opalmer:
Makes sense to me, merging.

PR #272, opened 2015-04-22
Allow jobtype implementors to set task progress

guidow:

PR #275, opened 2015-04-30
Import missing symbol BAD_REQUEST

guidow:


guidow:
This should go into #272 instead

PR #276, opened 2015-05-06
Catch AccessDenied when checking pid on startup

guidow:
On Windows, when running as a non-privileged user, we sometimes fail to
start up if our pid file is stale, and the pid has since been taken by
another process running under some other user. This should fix that.


opalmer:
Looks good to me, merging.

PR #277, opened 2015-05-06
Configurably reject multiple simultaneous batches

guidow:
This makes agent sharing configurable. By default, all new assignments
will be rejected if we still have at least one assignment with at least
one task that is not done or failed.

This also slightly changes the semantics of the /assign endpoint:
The agent will now return 202 ACCEPTED if it is sent an assignment that
is identical to one we already have, but won't internally register and
start a new jobtype for it. There is now no longer a way for a client
to tell the difference between that and a fresh assignment, effectively
making the endpoint idempotent.

This is necessary because 409 CONFLICT is now a hard error that means
the assignment will NOT be executed and that the master will have to
react to. (Usually by finding a new node for the rejected task ids.)

### DIFF
--- a/pyfarm/agent/etc/agent.yml
+++ b/pyfarm/agent/etc/agent.yml
@@ -209,3 +209,5 @@ sysinfo_command_lspci:
   - /sbin/lspci
   - /usr/sbin/lspci
   - /usr/bin/lspci
+
+agent_allow_sharing: False
--- a/pyfarm/agent/http/api/assign.py
+++ b/pyfarm/agent/http/api/assign.py
@@ -145,20 +145,43 @@ class Assign(APIResource):
         except AttributeError:  # pragma: no cover
             current_assignments = config["current_assignments"].values
 
-        existing_task_ids = set()
         new_task_ids = set(task["id"] for task in request_data["tasks"])
 
         for assignment in current_assignments():
-            for task in assignment["tasks"]:
-                existing_task_ids.add(task["id"])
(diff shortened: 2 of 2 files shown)


pyfarm/agent/http/api/assign.py
```diff
@@ -145,20 +145,43 @@ def post(self, **kwargs):
...
-                existing_task_ids.add(task["id"])
-
-        if existing_task_ids & new_task_ids:
-            request.setResponseCode(CONFLICT)
-            request.write(
-                dumps({"error": "Double assignment of tasks",
-                       "duplicate_tasks": list(existing_task_ids & new_task_ids)}))
-            request.finish()
-            return NOT_DONE_YET
+            existing_task_ids = set([x["id"] for x in assignment["tasks"]])
```
opalmer: You don't need to construct the list before hand:

```python
set(x["id"] for x in assignment["tasks"])
```

pyfarm/agent/etc/agent.yml
```diff
@@ -209,3 +209,5 @@ sysinfo_command_lspci:
   - /sbin/lspci
   - /usr/sbin/lspci
   - /usr/bin/lspci
+
+agent_allow_sharing: False
```
opalmer: Can we get a doc string above this?  Copying part of what's in the description of this PR would probably be fine.

opalmer:
Design looks good, seems like a reasonable change to me.  Had a few code comment though but nothing major.  If you can fix this I don't see any reason this can't merge.

PR #278, opened 2015-05-07
On assign, check agent_id if provided

guidow:


pyfarm/agent/http/api/assign.py
```diff
@@ -68,6 +68,17 @@ def post(self, **kwargs):
         if request_from_master(request):
             config.master_contacted()
 
+        if ("agent_id" in request_data and
+            request_data["agent_id"] != config["agent_id"]):
+            logger.error("Wrong agent_id in assignemnt: %s. Our id is %s",
+                         request_data["agent_id"], config["agent_id"])
+            request.setResponseCode(BAD_REQUEST)
+            request.write(dumps(
+                {"error": "You have the wrong agent. I am %s." %
```
opalmer: I'd add an extra field to the error dump, `agent_id` which only has the id in it.  Makes it easier to extract the agent id and use it elsewhere.

opalmer:
Good call, only had one comment.  After that I think this can merge.

PR #280, opened 2015-05-08
Bugfixes and Tests for /tasklogs/

opalmer:
This change adds full coverage for the /tasklogs/ endpoint and also fixes a couple of latent bugs.

### DIFF
--- a/pyfarm/agent/http/api/tasklogs.py
+++ b/pyfarm/agent/http/api/tasklogs.py
@@ -82,7 +82,7 @@ class TaskLogs(APIResource):
         log_identifier = request.postpath[0]
         if "/" in log_identifier or "\\" in log_identifier:
             request.setResponseCode(BAD_REQUEST)
-            request.write(dumps({"error": "log_identifier must not contain"
+            request.write(dumps({"error": "log_identifier must not contain "
                                           "directory separators"}))
 
         path = join(config["jobtype_task_logs"], log_identifier)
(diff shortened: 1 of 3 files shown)


opalmer:
@guidow, PTAL.  I'm trying to shift my focus a bit to increasing coverage on the agent where we're missing it and this seemed like some low hanging fruit to hit first.  Eventually I want to get back to converting over to inline callbacks....but we need better tests and more coverage first I think.

pyfarm/agent/http/api/tasklogs.py
```diff
@@ -43,23 +87,41 @@ def get(self, **kwargs):
 
         if len(request.postpath) != 1:
             request.setResponseCode(BAD_REQUEST)
+            request.setHeader("Content-Type", "application/json")
```
guidow: This is wrong. The content is text/csv.

tests/test_agent/test_http_api_tasklogs.py
```diff
@@ -0,0 +1,184 @@
...
+        with open(self.log, "wb") as log:
+            log.write(self.log_content)
+
+    def test_methods_exist_for_schema(self):
+        self.skipTest("No schemas defined")
+
+    def test_master_contacted(self):
+        """
+        Make sure that if the request is coming
+        from the master that is sets the value in
```
guidow: Typo, maybe some rephrasing needed

guidow:
Looks good

PR #281, opened 2015-05-08
remove outdated and unused jobtypes examples

opalmer:
The examples.py file has been updated for quite some time.  Additionally this is something that probably belongs in the docs rather than inside of the code base itself anyway.


opalmer:
@guidow, PTAL

guidow:
As simple as this example is, I think it would actually still work.

If you think this doesn't belong here, though, go ahead and remove it.

opalmer:
Yeah I think I'll go ahead and remove it.  We really need a better set of examples for one and they should also go somewhere more visible such as the docs...

PR #282, opened 2015-05-11
Missing Content-Type or Bad Codes In Some API Responses

opalmer:
If the return value from a resource was either a string or a two part tuple the default content type header
was not being set.   The return value from `get_content_type()` was also incorrect (should be a list so setRawHeaders() can use it).  There were also cases where the code could be set incorrectly if both the rendering function and the base resource try to set the code.

### DIFF
--- a/pyfarm/agent/http/core/resource.py
+++ b/pyfarm/agent/http/core/resource.py
@@ -143,7 +143,7 @@ class Resource(_Resource):
         """
         header = request.requestHeaders.getRawHeaders("Content-Type")
         if not header:
-            return list(self.DEFAULT_CONTENT_TYPE)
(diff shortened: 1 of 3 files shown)


opalmer:
@guidow, PTAL.  Found this bug while writing tests for tasks and tasklogs.

guidow:
Looks okay so far.

opalmer:
Have you had a chance to run this yet?  If so and you haven't had any problems we should probably go ahead and merge this.

guidow:
Yes, I've had a chance to run this in our test farm, and I haven't seen anything amiss.

opalmer:
Cool thanks, I'll go ahead and merge this then.  Just wanted to be sure.

PR #284, opened 2015-05-11
Reject assignments if we are busy reannouncing

guidow:
Accepting assignments during a running reannounce can lead to weird
race, where the reannounce request is locally started right before the
incoming assignment, and uses data from that point in time, but doesn't
actually reach the master until after the incoming assignment. At that
point, the master has already marked the tasks as sent to agent and will
assume we failed it because our announce is lacking those task ids.

### DIFF
--- a/pyfarm/agent/http/api/assign.py
+++ b/pyfarm/agent/http/api/assign.py
@@ -81,7 +81,7 @@ class Assign(APIResource):
             request.finish()
             return NOT_DONE_YET
 
-        if config["agent"].reannounce_lock.locked:
+        if self.agent.reannounce_lock.locked:
             logger.warning("Temporarily rejecting assignment because we "
                            "are in the middle of a reannounce.")
             request.setResponseCode(BAD_REQUEST)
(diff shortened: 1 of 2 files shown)


pyfarm/agent/http/api/assign.py
```diff
@@ -81,6 +81,16 @@ def post(self, **kwargs):
...
 
+        if self.agent.reannounce_lock.locked:
+            logger.warning("Temporarily rejecting assignment because we "
+                           "are in the middle of a reannounce.")
+            request.setResponseCode(BAD_REQUEST)
+            request.write(
+                dumps({"error": "Agent cannot accept assignments because of a "
+                                "reannounce in progress. Try again shortly."}))
+            request.finish()
+            return NOT_DONE_YET
```
opalmer: It would be cleaner and more efficient to just return the body and response code:

```python
return dumps({"error": "Agent cannot accept assignments because of a "
                       "reannounce in progress. Try again shortly."}), BAD_REQUEST
```

However, there's a bug in the base resource that would essentially prevent `BAD_REQUEST` from being set.  #282 should fix this bug though.  I don't think it's worth holding back this change till that fixed is merge though given what this PR is fixing.

PR #285, opened 2015-05-12
Remember local state for all tasks in assignment

guidow:
This way, the master will see what state we think the assignment is in
when polling the agent.

### DIFF
--- a/pyfarm/jobtypes/core/jobtype.py
+++ b/pyfarm/jobtypes/core/jobtype.py
@@ -932,12 +932,14 @@ class JobType(Cache, System, Process, TypeChecks):
             logger.error("Not setting task to failed: agent is shutting down.")
 
         else:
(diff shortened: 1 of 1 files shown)


pyfarm/jobtypes/core/jobtype.py
```diff
@@ -932,6 +932,13 @@ def set_task_state(self, task, state, error=None, dissociate_agent=False):
             logger.error("Not setting task to failed: agent is shutting down.")
 
         else:
+            assignment_id = self.assignment["id"]
+            assignment = config["current_assignments"].get(assignment_id, None)
+            if assignment:
+                task_ = next(x for x in assignment["tasks"] if
```
opalmer: Hmm, I think a straight for loop with a break would be just as efficient but more readable.  `x` should also be named something more human friendly too.

opalmer:
Looks good, thanks

PR #286, opened 2015-05-13
Set status to SERVICE UNAVAILABLE for reannounce

guidow:
This status code fits the situation better. After all, it isn't the client's fault, and technically there was nothing wrong with the request, we just cannot service it at the moment.


opalmer:
Makes sense to me.  I actually had a similar thought when this code was first added, just couldn't come up with a better code at the time...

PR #288, opened 2015-06-02
Add JobType.set_task_started_now() method

guidow:


pyfarm/jobtypes/core/jobtype.py
```diff
@@ -890,6 +890,96 @@ def set_task_progress(self, task, progress):
...
+
+                    deferred = Deferred()
+                    reactor.callLater(delay, deferred.callback, None)
+                    yield deferred
+
+                elif response.code == NOT_FOUND:
+                    message = ("Got 404 NOT FOUND error on setting start time "
+                               "for task %s" % task["id"])
+                    logger.error(message)
+                    raise Exception(message)
```
opalmer: Perhaps we should make this a custom exception, similar to what we did with `InsufficientDiskSpace`, so it can be more explicitly handled?

pyfarm/jobtypes/core/jobtype.py
```diff
@@ -890,6 +890,96 @@ def set_task_progress(self, task, progress):
...
+
+                elif response.code >= BAD_REQUEST:
+                    message = (
+                        "Failed to set start time for task %s on the "
+                        "master, bad request: %s. Server.  This request will "
+                        "not be retried." % (task["id"], response_data))
+                    logger.error(message)
+                    raise Exception(message)
```
opalmer: Perhaps we should make this a custom exception, similar to what we did with `InsufficientDiskSpace`, so it can be more explicitly handled?

pyfarm/jobtypes/core/jobtype.py
```diff
@@ -890,6 +890,96 @@ def set_task_progress(self, task, progress):
...
+
+                else:
+                    message = (
+                        "Unhandled error when setting start time for task "
+                        "%s to the master: %s (code: %s).  This request will "
+                        "not be retried." % (response_data, response.code))
+                    logger.error(message)
+                    raise Exception(message)
```
opalmer: Perhaps we should make this a custom exception, similar to what we did with `InsufficientDiskSpace`, so it can be more explicitly handled?

pyfarm/jobtypes/core/jobtype.py
```diff
@@ -890,6 +890,96 @@ def set_task_progress(self, task, progress):
...
+
+            except (ResponseNeverReceived, RequestTransmissionFailed) as error:
+                num_retry_errors += 1
+                if num_retry_errors > config["broken_connection_max_retry"]:
+                    message = (
+                        "Failed to set start time for task %s to the "
+                        "master, caught try-again type errors %s times in a "
+                        "row." % (task["id"], num_retry_errors))
+                    logger.error(message)
+                    raise Exception(message)
```
opalmer: Perhaps we should make this a custom exception, similar to what we did with `InsufficientDiskSpace`, so it can be more explicitly handled?

pyfarm/jobtypes/core/jobtype.py
```diff
@@ -890,6 +890,96 @@ def set_task_progress(self, task, progress):
...
+        if not isinstance(task, dict):
+            raise TypeError(
+                "Expected a dictionary for `task`, cannot set start time")
+
+        if "id" not in task or not isinstance(task["id"], INTEGER_TYPES):
+            raise TypeError(
+                "Expected to find 'id' in `task` or for `task['id']` to "
+                "be an integer.")
+
+        url = "%s/jobs/%s/tasks/%s" % (
```
opalmer: nit, use string formatting for readability:

```python
url = "{master_api}/jobs/{job_id}/{task_id}".format(...)
```

pyfarm/jobtypes/core/jobtype.py
```diff
@@ -890,6 +890,96 @@ def set_task_progress(self, task, progress):
                     "%r." % (task["id"], error))
                 raise
 
+    @inlineCallbacks
+    def set_task_started_now(self, task):
```
opalmer: Hmm, is there any reason we don't want to call this `set_task_time_started` so you could specify a specific time (including 'now').  I know the usual behavior should be to set `time_started` to now but it may also be useful to set specific times as well.

pyfarm/jobtypes/core/jobtype.py
```diff
@@ -890,6 +890,96 @@ def set_task_progress(self, task, progress):
...
+        """
+        Sets the time_started of the given task to the current time on the
+        master.
+        This method is useful for batched tasks, where the actual work on a
```
opalmer: newline for a new paragraph please

pyfarm/jobtypes/core/jobtype.py
```diff
@@ -890,6 +890,96 @@ def set_task_progress(self, task, progress):
...
+        single task may start much later than the work the assignment as a
+        whole.
+
+        :param dict task:
+            The dictionary containing the task we're changing the start time
```
opalmer: nit:

`The dictionary containing the task we're changing ``start_time`` for`

pyfarm/jobtypes/core/jobtype.py
guidow: I cannot think of anything better than ```Exception``` for this case. ```BAD_REQUEST``` is a fairly generic error code, after all.

pyfarm/jobtypes/core/jobtype.py
guidow: Similar note as above, this is a generic catch-all handler, I cannot think of a better fitting exception name here.

pyfarm/jobtypes/core/jobtype.py
guidow: We have been using the % operator for string formatting just about everywhere in the code until now. Do you think we should switch over to string.format() in general?

pyfarm/jobtypes/core/jobtype.py
guidow: I was hoping to leave picking any actual timestamps to the master (so we have a consistent time source). Also, I don't really see a use-case for setting any other times.

opalmer: The use case I was thinking of is if you had a known delay when starting a job on a host.  But, I see your point from the master's perspective.  Withdrawn.

opalmer:
Alright I think we can go ahead and merge this.

PR #290, opened 2015-06-03
Don't convert jobtype name, job title to str

guidow:
This will fail if either of them contain non-ASCII characters.


opalmer:
Thanks, not sure exactly why that was being doing in the first place.

PR #291, opened 2015-06-03
Reduce severity of some log statements

guidow:
and substitute some more readable ones.

### DIFF
--- a/pyfarm/jobtypes/core/jobtype.py
+++ b/pyfarm/jobtypes/core/jobtype.py
@@ -1137,7 +1137,12 @@ class JobType(Cache, System, Process, TypeChecks):
             which contains the protocol used to communicate between the
             process and this job type.
         """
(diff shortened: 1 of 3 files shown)


pyfarm/jobtypes/core/jobtype.py
```diff
@@ -1137,7 +1137,12 @@ def before_spawn_process(self, command, protocol):
             which contains the protocol used to communicate between the
             process and this job type.
         """
-        logger.info("Spawning %r", command)
+        command_line = command.command
+        for argument in command.arguments:
```
opalmer: Little cleaner:

```python
" ".join([command.comand] + command.arguments)
```

guidow: You're right, that looks cleaner.

opalmer:
Looks good, merging.

PR #294, opened 2015-06-05
Send available disk space information to master

guidow:
Warning: This needs an accompanying change in the master, otherwise there will be an incompatibility.

### DIFF
--- a/pyfarm/agent/sysinfo/disks.py
+++ b/pyfarm/agent/sysinfo/disks.py
@@ -21,58 +21,28 @@ Disks
 Contains information about the local disks.
 """
 
-try:
-    from os import statvfs
-except ImportError:  # pragma: no cover
-    statvfs = NotImplemented
+import psutil
 
-try:
-    from wmi import WMI
-except ImportError:  # pragma: no cover
-    WMI = NotImplemented
-
-from pyfarm.core.enums import WINDOWS, LINUX, BSD
 from pyfarm.agent.logger import getLogger
+
+from collections import namedtuple
+DiskInfo = namedtuple("DiskInfo", ("mountpoint", "free", "size"))
 
 logger = getLogger("agent.disks")
 
 
-class DiskLookupError(Exception):
-    pass
-
-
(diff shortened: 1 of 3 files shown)


guidow:
Unfortunately, getting a list of disks available is something highly platform dependent again, and I have no OS X or BSD boxes to test this on.

pyfarm/agent/sysinfo/disks.py
```diff
@@ -0,0 +1,78 @@
...
+from pyfarm.agent.logger import getLogger
+
+logger = getLogger("agent.disks")
+
+
+class DiskLookupError(Exception):
+    pass
+
+
+class DiskInfo:
```
opalmer: We should used `namedtuple` for this.

pyfarm/agent/sysinfo/disks.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, mountpoint, free, size):
+        self.mountpoint = mountpoint
+        self.free = free
+        self.size = size
+
+
+def disks():
```
opalmer: Can we use `psutil` instead here since it's cross-platform and will be less work?  The below for example won't account for bind mounts and may also have trouble on Linux for some non-standard setups.

```python
from collections import namedtuple
DiskInfo = namedtuple("DiskInfo", ("mountpoint", "free", "size"))

for partition in psutil.disk_partitions():
    usage = psutil.disk_usage(partition.mountpoint)
    
    info = DiskInfo(
         mountpoint=usage.mountpoint,
         free=usage.free,
         size=usage.total
    )
    
```

guidow: That's a much better way, actually. I didn't think of psutil when I wrote this.

opalmer: Yeah the disk stuff is pretty new compared to other APIs provided by psutil.  It's certainly a nice alternative in this case.

opalmer:
Looks good, thanks.

PR #296, opened 2015-06-05
Fix getting hostname

guidow:


pyfarm/agent/sysinfo/network.py
```diff
@@ -120,15 +130,12 @@ def hostname(trust_name_from_ips=True):
     # produce the wrong information when resolving the local
     # hostname.
     if len(reverse_hostnames) == 1 and trust_name_from_ips:
-        return reverse_hostnames.pop()
+        return reverse_hostnames.pop().lower()
```
opalmer: Do we still need the lower here since the hostname should have been inserted after calling `.lower()`?  If yes, we probably are missing a `.lower()` call somewhere.

pyfarm/agent/sysinfo/network.py
```diff
@@ -93,11 +93,21 @@ def hostname(trust_name_from_ips=True):
     """
     logger.debug("Attempting to discover the hostname")
 
+    ip_addresses = addresses()
+
+    local_fqdn_query = socket.getfqdn().lower()
+    _h, _a, ips_from_fqdn = socket.gethostbyname_ex(local_fqdn_query)
```
opalmer: Probably need to wrap `socket.gethostbyname_ex` a bit so we ignore the most common case of a name the system still can't resolve (I've seen this happen before on misconfigured systems):

```python
try:
    _h, _a, ips_from_fqdn = socket.gethostbyname_ex(local_fqdn_query)
except socket.gaierror as error:
    if error.errno != ENOEXEC:
        raise
    ips_from_fqdn = set()
```

pyfarm/agent/sysinfo/network.py
guidow: No, that one was unnecessary.

pyfarm/agent/sysinfo/network.py
guidow: Why look specifically for ```ENOEXEC```?

opalmer: That's actually the wrong error code....it should come from the socket module.  Anyway I looked around and it looks like socket.gaierror does not have specific enough errors to have anymore specific behavior than just logging it (so nevermind on the if statement).

opalmer:
Ok loos good now, thanks.

PR #298, opened 2015-06-22
Fix creating error page

guidow:
Twisted-web is quite adamant about not wanting to write unicode objects.


opalmer:
Yeah...I had the same problem, still trying to figure out why they're forcing the encoding to non-unicode...

PR #299, opened 2015-06-22
Fix getting logged in users with newer psutil

guidow:
psutil 3.0 removed psutil.get_users() in favor of psutil.users().


guidow:
Merging directly, since it's a simple bugfix.

opalmer:
Thanks, looks good.

PR #300, opened 2015-07-27
Set task progress to 0.0 when started

guidow:


opalmer:
Looks good to me, merging.

PR #302, opened 2015-08-21
Do not treat insufficient temp space as failure

guidow:
When a task fails because we couldn't free enough disk space for it,
that doesn't mean there is anything inherently wrong with that task, it
only mean that task cannot be run on this node at this point in time.

If we only set the tasks state to queued and dissociate it with this
agent, however, there is a high likelihood that the master's scheduler
will just immediately reassign that task to us.  To avoid that, we add
this agent to the failed_on_agents list of that task.

### DIFF
--- a/pyfarm/jobtypes/core/jobtype.py
+++ b/pyfarm/jobtypes/core/jobtype.py
@@ -709,7 +709,8 @@ class JobType(Cache, System, Process, TypeChecks):
             command.set_default_environment(environment)
             self._spawn_process(command)
 
-    def stop(self, assignment_failed=False, error=None, signal="KILL"):
+    def stop(self, assignment_failed=False, avoid_reassignment=False,
+             error=None,  signal="KILL"):
         """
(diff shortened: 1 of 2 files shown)


pyfarm/jobtypes/core/jobtype.py
```diff
@@ -899,6 +910,98 @@ def set_task_progress(self, task, progress):
...
+            config["master_api"], self.assignment["job"]["id"], task["id"])
+        data = {"id": config["agent_id"]}
+
+        added = False
+        num_retry_errors = 0
+        while not added:
+            try:
+                response = yield post_direct(url, data=data)
+                response_data = yield treq.json_content(response)
+                if response.code in [OK, CREATED]:
```
opalmer: tuples are generally cheaper

opalmer:
Looks good, saw a couple of places that could be improved but nothing major or worth making different than a few other places in the code base.  Merging.

PR #303, opened 2015-08-25
Windows Testing With Appveyor

opalmer:
This PR bring Windows testing to pyfarm's agent via the Appveyor service.  Just like the Travis tests the CI service will be used to sign-off on a PR.  This PR also has the changes necessary to get the tests passing on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,6 @@
 build
 dist
 MANIFEST
-_trial_temp
+_trial_temp*
 *.orig
 htmlcov
--- a/README.rst
+++ b/README.rst
@@ -20,7 +20,7 @@ PyFarm Agent
     :alt: build status (agent) (posix)
 
 .. image:: https://ci.appveyor.com/api/projects/status/a0fwqwlqrcs57sfn/branch/master?svg=true
-    :target: https://ci.appveyor.com/project/pyfarm/pyfarm-agent/history
+    :target: https://ci.appveyor.com/project/opalmer/pyfarm-agent/history
     :alt: build status (agent) (windows)
 
 .. image:: https://coveralls.io/repos/pyfarm/pyfarm-agent/badge.png?branch=master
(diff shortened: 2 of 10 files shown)


opalmer:
@guidow, PTAL.  There's still a little work to do on this PR but it's worth starting the review process:

* Review an merge the other PRs for the agent.  Many of those will fix tests which don't work on Windows.
* Fix the broken Python 2.6 test runner for Appveyor or remove it.  Appveyor does not fully support Python 2.6 like they do 2.7 which comes preinstalled and setup to properly compile libraries.  We'd still test 2.6 on Travis of course.
* Optionally, add SSL testing like we have on Travis.  Like the Python 2.6 issue we might just want to make Travis our primary testing platform for this and not do this level of testing using Appveyor. 

The last two points, if we decided to do them, could be a separate PR after this merges too.

guidow:
Apart from the obvious (some of the tests still failing) I don't see anything objectionable in the code.

> Fix the broken Python 2.6 test runner for Appveyor or remove it. Appveyor does not fully support Python 2.6 like they do 2.7 which comes preinstalled and setup to properly compile libraries. We'd still test 2.6 on Travis of course.

Personally, I see very little reason to keep testing with 2.6. 2.7 has been out for many years now.

> Optionally, add SSL testing like we have on Travis.

Hm, at the moment the moment, pyfarm isn't using SSL in practice anyway. Security-wise, pyfarm is still wide open on all fronts for all intents and purposes, and the only sensible way to run it is on a heavily firewalled network.
OTOH, if we do support SSL, we should probably test it on all platforms, at least in the long term. It's no big deal, IMHO, if the tests aren't in place at this point in time.

opalmer:
> Personally, I see very little reason to keep testing with 2.6. 2.7 has been out for many years now.

As much as I'd like to stop supporting 2.6, RHEL 6 still has Python 2.6 as the default and CentOS does as well I believe.  I'd say the cutoff for 2.6 support should be either when RHEL moves on or when Twisted supports 3.0 (whichever comes first).

opalmer:
Going to go ahead and merge this.  The test skips that I added can be removed in other PRs.

guidow:
RHEL/CentOS 6 will never move away from Python 2.6. It will likely only get security fixes and major bug fixes until it reaches EOL. You will have to go with RHEL 7 to get Python 2.7 by default.

opalmer:
Yeah good point, even Debian based distros don't upgrade the core Python version.  Not sure what I was thinking there, I think I was under the mistaken assumption that RHEL 7 hand't been released yet.  Sorry my bad.

I still think we should support 2.6 until for a little longer.  I'd like the support it as long as Twisted still does or until they release 3.0 support.  At the very least we should support 2.7 as long as we're not testing across all three platforms (which TBH, won't be much longer).

PR #305, opened 2015-08-25
Cleanup log.py some and improve Windows handling

opalmer:
This PR cleans up the internals of flush some and also ensures that we handle `WindowsError` when it's available.  Found most of these improvements while working #303, figured they could be broken out.


opalmer:
@guidow, PTAL

guidow:
Looks okay to me.

PR #306, opened 2015-08-26
disks.disks() should ignore OSError

opalmer:
From a comment in the code explaining this:

```
Not all disks can return disk information.  A partition
that is mounted but does not have a file system, cdrom
drives on Windows for example, wouldn't have any usage
data to return.
```

Found this problem on Windows while working on #303.


opalmer:
@guidow, PTAL

guidow:
Looks okay to me.

PR #307, opened 2015-08-26
[WIP] Make communication results from the master easier to test and introspect

opalmer:
While working on #303 it became more apparent that the results from functions which communicate with the master are not very descriptive of what happened.  They often return either None or a dictionary containing the response.  In the unittests this means we often have to mock logging calls rather than directly checking on the results.  For debugging by hand you have to look at the logs in addition to the return value when the return value alone could just be more descriptive.


opalmer:
And after diving into this more...this turned out to be a bad idea.  It's create additional overhead and wouldn't be much easier to setup that what we already have for testing.

PR #309, opened 2015-08-28
Improve Usage of disks.disks()

opalmer:
Since we're catching OSError inside of disks.disks() we shouldn't need the bare try/except clause anymore.  This PR also adds the tests for disks.disks() and provides a `as_dict` keyword so we can skip the extra for loop in service.py


opalmer:
@guidow, PTAL

guidow:
Looks good to me.

PR #310, opened 2015-08-30
Remove warning about PyOpenSSL in setup.py on Windows

opalmer:
Comment on ea11331 should explain most of the reasoning behind this.  I decided to remove the warning after seeing that on Windows this warning is usually shown to a user several times.  Additionally you can't really tell which package the warning came from which is misleading and could lead someone to believe something is wrong when there probably isn't.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,13 +16,12 @@
 
 from __future__ import with_statement
 
-import sys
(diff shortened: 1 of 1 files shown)


opalmer:
@guidow, PTAL

guidow:
Good call.

PR #311, opened 2015-08-30
Lower some of the logging calls in core/log.py to DEBUG

opalmer:
The log emissions in this change should be DEBUG rather than
INFO.  Either a jobtype would be logging the file path, the
information we're logging is not useful outside of debug/testing or
there are other way we'd find out there's a problem (such as an
exception being raised).


opalmer:
@guidow, PTAL

guidow:
Looks okay to me.

PR #312, opened 2015-08-30
Capture Logging/Silence tests in setUp

opalmer:
For comparison, here's the new output: https://travis-ci.org/pyfarm/pyfarm-agent/jobs/77930978
And here's the old output: https://travis-ci.org/pyfarm/pyfarm-agent/jobs/77850694

There's still a couple of places we're emitting output, but those will not be as easy to capture because of the context in which the logger is being used.


opalmer:
@guidow, PTAL

guidow:
Looks good.

PR #315, opened 2015-09-01
Convert test_core_process to use @inlineCallbacks

opalmer:
This is a minor enhancement to test_core_process.py to use inlineCallbacks.  The test flow is more linear and has fewer indirect calls to the test functions.

This PR also improves on the internals of TestStopProcess too.


opalmer:
@guidow, PTAL

guidow:
No objections from me.

PR #317, opened 2015-09-02
OSX Testing With Travis

opalmer:
As the title says, let's run our tests on OS X.


opalmer:
@guidow, PTAL this should bring us testing on OS X now too.

guidow:
Is it not possible to set the python versions to test against in the same way for OS X as for Linux?

Other than that, looks good.

opalmer:
> Is it not possible to set the python versions to test against in the same way for OS X as for Linux?

Nope, not according to https://github.com/travis-ci/travis-ci/issues/2312.

PR #318, opened 2015-09-02
Remove HTTPs Testing

opalmer:
Remove the extra https tests on the agent for now.  SSL is not being utilized yet and we're not consistently testing this across platforms so it does not make sense to keep it around at the moment.

This is a follow-on to part of the discussion in #303.


opalmer:
@guidow, PTAL.

opalmer:
I'm going to go ahead and close this.  It makes more sense to remove these tests in #317 since the yaml file for Travis is changing.

PR #319, opened 2015-09-02
Fix TestStatus.test_get_result

opalmer:


opalmer:
I'm going to go ahead and merge this since it fixes a test that's broken in other PRs.

guidow:
BTW, it might be a good idea to turn off maxDiff by default. When there's a failing test because some assertEqual failed, chances are pretty good we will want to see the whole diff.

opalmer:
Yep was doing that tonight: 24ace5d1368169f85dab73d5ae9499cd37c365ac.

PR #321, opened 2015-09-04
Don't repeat requests on 4xx errors

guidow:
For some time now, the error handler for 4xx errors when setting the
state for a task will call the errback on its own deferred.  This is
needed so we can remove this task deferred from the jobtype's
task_update_deferreds.  Unfortunately, the default error_callback will
always retry the failed requests.  This oversight can lead an agent to
keep trying to update its tasks on the master endlessly, even when the
master will reject the update.

This commit should fix that.


guidow:
Tested, nope, that wasn't it...

PR #322, opened 2015-09-08
Endless retry fix

guidow:


opalmer:
Looks good, thanks for catching this and for the inlineCallbacks conversion...much cleaner than the old implementation.

PR #325, opened 2015-09-21
Don't delay restart for a non-running assignment

guidow:
The previous code could get the agent stuck in an endless loop, because
the code that checks whether the agent should restart comes before the
code that removes the jobtype from config


opalmer:
Looks good, merging.

PR #326, opened 2015-09-21
Add method in JobType to clean up system temp dir

guidow:
We are having problems with render nodes whose temp directories are filling up. This should offer a way to mitigate those problems.

### DIFF
--- a/pyfarm/jobtypes/core/internals.py
+++ b/pyfarm/jobtypes/core/internals.py
@@ -786,6 +786,36 @@ class System(object):
         reactor.callInThread(self._remove_directories, self._tempdirs.copy())
         self._tempdirs.clear()
 
+    def _cleanup_system_temp_dir(self, minimum_age=604800):
+        """
+        Cleans up old file from the system's temp directory to reclaim space.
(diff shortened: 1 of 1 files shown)


opalmer:
Makes sense to me, no major objections since it's simple and not a public method (since a cron or other setup would be better).

The question I do have though is how do you see this being used?  Do you guys plan on using it all the time or at an interval somehow?  The reason I ask is because I'm wondering of this is something we should integrate at the service level and enable with a command line flag.  It would be more optimal there and would be something a job-type implementor wouldn't have to worry about at that point.  Plus that would give a system administrator more control over a system wide(ish) action too which is generally always a good idea.

guidow:
At the moment, we call it from the jobtypes, analogous to _ensure_free_space_in_temp_dir().

I think it's important that this isn't called by default anywhere, since it deletes files from the system, so the user should have to make a conscious decision to use this.

opalmer:
> I think it's important that this isn't called by default anywhere, since it deletes files from the system, so the user should have to make a conscious decision to use this.

Agreed.  What I meant by my suggestion was have it something someone can optionally enable at the service level.

PR #327, opened 2015-09-22
Fix problems with coverage and some testing improvements

opalmer:
This PR fixes some issues with regards to testing and code coverage reporting:

* Gets trial/coverage working again as single command
* Adds a missing `pyasn1` package which we seems to be a missing requirement if we use setup.py without pip: [https://travis-ci.org/pyfarm/pyfarm-agent/jobs/81516920](some circumstances).  This package does not have any native extensions so it should be easy to acquire and install on all platforms.
* Installs the correct version of coverage which coveralls requires
* Installs the wheel package which will be a little faster in most cases
* Uses setup.py develop to install the package.  This is a little cleaner and also fixes the path we're reporting on coveralls.io: [old](https://coveralls.io/builds/3609766) | [new](https://coveralls.io/jobs/8452597)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -26,3 +26,4 @@ omit =
     *pyfarm/agent/entrypoints/development.py*
     *pyfarm/agent/entrypoints/main.py*
     *pyfarm/agent/entrypoints/supervisor.py*
+    /home/*/virtualenv*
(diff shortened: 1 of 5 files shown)


opalmer:
@guidow, PTAL

guidow:
Looks good. I like the idea of running trial as just one command, this should save us the hassle of having to look for errors somewhere in the middle of a very large log file.

opalmer:
Thanks, couldn't agree more....We were doing it the old way I think because we didn't have \_\_init\_\_.py files in the tests directories (so trial didn't know what to test).

PR #328, opened 2015-09-22
Fix bug in JobType._cleanup_system_temp_dir()

guidow:


opalmer:
Oops my bad, should have noticed that in the review.  Merging.

PR #329, opened 2015-09-22
Catch broken connections when registering logfile

guidow:


pyfarm/jobtypes/core/internals.py
```diff
@@ -607,13 +609,30 @@ def result_callback(task, log_path, deferred, response):
...
-                log_path, task["id"], failure_reason, delay)
-            post_logfile(task, log_path, post_deferred=deferred, delay=delay)
+        def error_callback(task, log_path, deferred, num_retry_errors,
+                           failure_reason):
+            if num_retry_errors > config["broken_connection_max_retry"]:
+                logger.error(
+                    "Error while registering logfile %s for task %s on master. "
+                    "Maximum number of retries reached. Not retrying the "
+                    "request.", log_path, task["id"])
+                deferred.errback
```
opalmer: Did you mean to do something here with the errback?

pyfarm/jobtypes/core/internals.py
```diff
@@ -607,13 +609,30 @@ def result_callback(task, log_path, deferred, response):
...
+                    logger.debug(
+                        "Error while registering logfile %s for task %s on "
+                        "master: %s, retrying immediately",
+                        log_path, task["id"], failure_reason.type.__name__)
+                    post_logfile(task, log_path, post_deferred=deferred)
+                else:
+                    delay = http_retry_delay()
+                    logger.error(
+                        "Error while registering logfile %s for task %s on "
+                        "master: \"%s\", retrying in %s seconds.",
```
opalmer: Use `%r` so you don't have to escape the quotes.  It will also render properly if it happens to not be a string for some reason.

opalmer:
Generally looks fine, had a couple of comments than this can merge I think.  I look forward to when we're using inlineCallbacks for these...

pyfarm/jobtypes/core/internals.py
guidow: Yes, I meant to call it...

PR #330, opened 2015-09-22
Fix restart API endpoint

guidow:
jobtypes was misspelled and the endpoint would fail when there was no
"jobtypes" in config yet.

### DIFF
--- a/pyfarm/agent/http/api/state.py
+++ b/pyfarm/agent/http/api/state.py
@@ -109,7 +109,8 @@ class Restart(APIResource):
                 logger.info("Created run control file %s",
                             config["run_control_file"])
 
-        if not config["jobtype"] or data.get("immediately", False):
+        if ("jobtypes" not in config or not config["jobtypes"] or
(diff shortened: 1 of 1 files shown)


pyfarm/agent/http/api/state.py
```diff
@@ -109,7 +109,8 @@ def post(self, **kwargs):
                 logger.info("Created run control file %s",
                             config["run_control_file"])
 
-        if not config["jobtype"] or data.get("immediately", False):
+        if ("jobtypes" not in config or not config["jobtypes"] or
```
opalmer: This would probably be simpler:

```python
if config.get("jobtypes") is None or data.get("immediately", False)
```

guidow: Hm, no. I don't just want to catch when it is ```None```, I also need to check if it's an empty list.

opalmer: Alright, makes sense thanks.

PR #331, opened 2015-09-23
Basic tests for sysinfo.graphics.graphics_cards()

opalmer:


opalmer:
@guidow, ptal.  I thought about mocking the Popen and WMI calls but since we're looking for some very specific output, which varies by platform, I though we'd go with this instead.  I'd welcome some input if you had other ideas though.

opalmer:
Alright, I think this is ready now.  @guidow PTAL, these updates should give us full coverage of `sysinfo.graphics` regardless of where the tests are executing.

guidow:
Looks okay to me.

PR #333, opened 2015-09-24
Make test_cleans_up_disk_space() more reliable

opalmer:
This particular test has been a source of many of our recent failures, this PR should fix that.

From the commit message body:
```
Depending on the operating system, state of the disk
and other factors it's more reliable to check how many
files we cleaned up rather than explicit list.  In the real
world, we care more about the quantity/size anyway rather
than specific file names.
```


opalmer:
@guidow, PTAL

guidow:
Looks good, thanks for writing the test.

PR #334, opened 2015-09-27
[WIP] JobTypeLoader Enhancement

opalmer:
This PR provides a few bugfixes, changes and improvements to the way we're loading job types.  Overall the code is simpler than the old setup and more thoroughly tested as well.  Here's a complete list of changes/enhancements coming out of this PR over the old design:

**Bugfixes**

* If a job type was removed from the master between the time we received work and when we started, the download request will not fail.  Before continue to try and process the result anyway.
* We now explicitly handle individual http codes from the master.  Before the code assumed any http response >= 500 should be retried and everything else was ok.

**Refactors**

* `Cache` has been renamed to `JobTypeLoader`.  This new name better fits the intended function and is something that is intended to perform the entire process of retrieving and returning a job type rather than mostly just caching.
* `JobTypeLoader` is no longer a parent class of `JobType`.  Instead it's instanced as the class variable `JobType.LOADER`.  This simplifies the overall class structure some, puts fewer inherited methods under each JobType instance and also allows a JobType implementor (or tests) to add their own loader more easily.

**Enhancements**
* JobType.load() has been simplified. Now it only validates the schema then hands off the whole loading process to ``JobTypeLoader``.
* Caching is disk based only now.  We were not getting many performance benefits with disk based and ram based caching.  The old setup was also more complicated than it probably needed to be too.  This should also reduce the possibility of chewing through extra ram for larger job types.
* Unlike `Cache`, `JobTypeLoader` can be tested and used independently of a JobType instance more easily.

### DIFF
--- a/pyfarm/jobtypes/core/internals.py
+++ b/pyfarm/jobtypes/core/internals.py
@@ -46,15 +46,18 @@ except ImportError:  # pragma: no cover
 
 try:
     from httplib import (
-        OK, INTERNAL_SERVER_ERROR, CREATED, ACCEPTED, CONFLICT)
+        OK, INTERNAL_SERVER_ERROR, CREATED, ACCEPTED, CONFLICT, NOT_FOUND,
+        BAD_REQUEST)
 except ImportError:  # pragma: no cover
     from http.client import (
-        OK, INTERNAL_SERVER_ERROR, CREATED, ACCEPTED, CONFLICT)
+        OK, INTERNAL_SERVER_ERROR, CREATED, ACCEPTED, CONFLICT, NOT_FOUND,
+        BAD_REQUEST)
 
 from psutil import disk_usage
 
 from twisted.internet import reactor, threads
-from twisted.internet.defer import Deferred, DeferredList, succeed
+from twisted.internet.defer import (
+    Deferred, DeferredList, succeed, inlineCallbacks, returnValue)
 from twisted.web._newclient import (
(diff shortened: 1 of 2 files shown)


pyfarm/jobtypes/core/internals.py
```diff
@@ -83,88 +87,222 @@ class InsufficientSpaceError(Exception):
...
+
+class JobTypeLoader(object):
+    """Class for retrieval, loading and caching of job types."""
+    def __init__(self):
+        if not config["jobtype_enable_cache"]:
+            cache_directory = None
         else:
-            logger.info("Created job type cache directory %r", CACHE_DIRECTORY)
+            try:
+                cache_directory = config["jobtype_cache_directory"]
```
guidow: Could we join ```config['farm_name']``` to this, if it is set?

That way, if agents get switched between farms for some reason, they won't get the different jobtypes caches mixed up.

opalmer: Yes fixed, good catch thanks.

opalmer:
For reference, still a work in progress right now.  There are either some issues with the `TestAssign.test_accepted` test or something in `Assign.post` itself is not playing well with inlineCallbacks.  Right now my theory is that because test_accepted does not wait for a result and is more or less executed in a linear fashion it's causing issues and not allowing JobTypeLoader to finish the work it needs.

PR #335, opened 2015-10-14
Modify assign API to respond directly with tuples

opalmer:
So while working on #334 it's become more apparent that in order to handle exceptions properly and generally cleanup the assignment APIs we need to do a little work converting it to inlineCallbacks.  Otherwise we can end up with a response going back to a client before a job type is actually loaded causing exceptions to possibly swallowed.

The modification of the assignment API to use inlineCallbacks will be broken down into another CL but for now this PR modifies some of the code from this style of response:

```python
request.setResponseCode(ACCEPTED)
request.write(dumps({"id": assignment_uuid}))
request.finish()
return NOT_DONE_YET
```

to:

```python
return dumps({"id": assignment_uuid}), ACCEPTED
```

This approach is cleaner and tightens the coupling some between the request being made and when the response is sent.

### DIFF
--- a/pyfarm/agent/http/api/assign.py
+++ b/pyfarm/agent/http/api/assign.py
@@ -193,12 +193,6 @@ class Assign(APIResource):
         assignment_uuid = uuid4()
         request_data.update(id=assignment_uuid)
         config["current_assignments"][assignment_uuid] = request_data
-
-        # In all other cases we have some work to do inside of
-        # deferreds so we just have to respond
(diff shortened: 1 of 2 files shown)


opalmer:
@guidow, PTAL

opalmer:
@guidow, any comments on this one?

guidow:
Looks good. Sorry for the long delay.

opalmer:
No problem, thanks for the comment.  Merging.

PR #336, opened 2015-10-23
Protect only $TMP/pyfarm from deletion

guidow:
As opposed to directory named "pyfarm" anywhere under $TMP

### DIFF
--- a/pyfarm/jobtypes/core/internals.py
+++ b/pyfarm/jobtypes/core/internals.py
@@ -825,7 +825,7 @@ class System(object):
         for root, dirs, files in os.walk(tempdir, topdown=True,
                                          followlinks=False):
             # Don't delete our own temp files
-            if "pyfarm" in dirs:
+            if root == tempdir and "pyfarm" in dirs:
                 dirs.remove("pyfarm")
             for filename in files:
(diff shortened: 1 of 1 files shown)


pyfarm/jobtypes/core/internals.py
```diff
@@ -825,7 +825,7 @@ def _cleanup_system_temp_dir(self, minimum_age=604800):
         for root, dirs, files in os.walk(tempdir, topdown=True,
                                          followlinks=False):
             # Don't delete our own temp files
-            if "pyfarm" in dirs:
+            if root == tempdir and "pyfarm" in dirs:
```
opalmer: The variable `config.tempdir` should be something like `/tmp/pyfarm/agent` so we could just use that to skip our root I think:

```python
# Don't delete our own temp files (agent or otherwise)
if root == dirname(config.tempdir):
    dirs[:] = []  # Do not iterate over sub directories in this root
    files[:] = []  # Do not iterate over files in this root
```

I think this will be more accurate than just checking for 'pyfarm' in dirs, especially if we the config changes.

opalmer:
Looks good thanks, merging.

PR #337, opened 2015-10-26
Fix test_sysinfo.TestNetwork.test_interfaces test

opalmer:
This test assumed there's always more than one interface which
may not always be the case.  For example you've shutoff all network
adapters you should be left with one interface alone.  This of course
assumes all hosts have at least the loopback interface.


opalmer:
@guidow, PTAL

guidow:
Looks okay.

PR #339, opened 2015-10-30
Reraise KeyError in LoggerPool.log with a better error message

opalmer:
KeyError can be raised if you don't call open_log() first.  This
error when running the tests is common for example.  open_log()
is usually called when the job type is started but if you're
writing tests or manually using the log pool the error message
could be more helpful.


opalmer:
@guidow, PTAL

guidow:
Looks good to me.

PR #340, opened 2015-11-01
JobType.tempdir() should handle IOError and WindowsError too

opalmer:
Like most calls to os.makedirs, we should catch IOError and WindowsError too as they can crop up from time to time.


opalmer:
@guidow, PTAL

guidow:
Looks good

PR #341, opened 2015-11-01
Simplify JobType.get_uid_gid() logic

opalmer:
This PR simplifies the logic inside of get_uid_gid() by:

* Removing the extra construction of a list
* Removing the unnecessary calls to all()
* Reworking the if statement a bit so it's faster


opalmer:
@guidow, PTAL

guidow:
Okay

PR #343, opened 2015-11-01
Fix/Improve JobType.get_csvlog_path()

opalmer:
This PR improves the logic we had which was doing a type check on the inputs to `get_csvlog_path`.  Before we could have thrown an AssertionError when we probably meant to be throwing TypeError instead via `_check_csvlog_path_inputs()`.


opalmer:
@guidow, PTAL

guidow:
Looks okay

PR #345, opened 2015-11-01
Fix JobType.get_environment()

opalmer:
This PR updates `get_environment()` a bit so it's more apparent when there's a problem with the configuration.  Before this you could accidentally provide a non-dictionary for `jobtype_default_environment` and you'd never know unless you looked at the agent logs.


opalmer:
@guidow, PTAL

PR #346, opened 2015-11-01
Fix comparison in handle_stderr_line() and normalize code with handle_stdout_line()

opalmer:


opalmer:
@guidow, PTAL

guidow:
Looks okay.

PR #347, opened 2015-11-01
Fix variable in JobType.log_stderr_line() and add tests

opalmer:


opalmer:
@guidow, PTAL

guidow:
Looks good to me.

PR #349, opened 2015-11-01
Enhancements and coverage for JobType.format_error()

opalmer:


opalmer:
@guidow, PTAL

guidow:
Looks good. Removing the special case for Exception and just calling str() on the object seems like a good idea.

You may need to rebase this branch to master.

PR #350, opened 2015-11-03
Psutil 3 Upgrade

opalmer:
This PR upgrade our `sysinfo.network` module to use Psutil 3.0.  By doing this we're also able to remove netifaces as a dependency.

### DIFF
--- a/pyfarm/agent/sysinfo/network.py
+++ b/pyfarm/agent/sysinfo/network.py
@@ -34,7 +34,6 @@ sent/received, and some error information.
 
 import socket
 
-import netifaces
 import psutil
 from netaddr import IPSet, IPNetwork, IPAddress
 
@@ -181,13 +180,11 @@ def addresses(private_only=True):
             if nic.address is None:
                 continue
 
-            if nic.family == socket.AF_INET:
+            elif nic.family == socket.AF_INET:
                 if not private_only:
                     results.add(nic.address)
                     continue
 
-                # Make sure that what we're getting out of
-                # netifaces is something we can use.
                 try:
                     if IPAddress(nic.address) in IP_PRIVATE:
                         results.add(nic.address)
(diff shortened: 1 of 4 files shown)


opalmer:
@guidow, PTAL

pyfarm/agent/sysinfo/network.py
```diff
@@ -171,23 +173,29 @@ def hostname(trust_name_from_ips=True):
...
-                        "Could not convert %s to a valid IP object" % addr)
-                else:
-                    if ip in IP_PRIVATE or not private_only:
-                        results.add(addr)
+                        "Could not convert %s to a valid IP object",
+                        nic.address)
+
+            elif nic.family == socket.AF_INET6:
+                logger.warning(
+                    "IPv6 not yet supported in addresses() for %s", name)
```
guidow: While I am still in favor of supporting IPv6 eventually, until we actually officially try to do so, we should get rid of this warning, IMHO. It just clutters the logs with urgent looking warnings without actually being important at all.

pyfarm/agent/sysinfo/network.py
```diff
@@ -66,18 +66,20 @@ def mac_addresses(long_addresses=False, as_integers=False):
...
+            if nic.family != psutil.AF_LINK:
+                continue
+
+            if not long_addresses and len(nic.address) > 17:
+                continue
+
+            mac = nic.address
+            if as_integers:
+                mac = int("0x" + nic.address.replace(":", ""), 0)
+            results.add(mac)
```
guidow: This code fails to filter out 00:00:00:00:00:00 mac addresses (from the loopback interface), as the old code did.

guidow:
I'm generally in favor of removing the dependency on netifaces, but see my other comments.

pyfarm/agent/sysinfo/network.py
opalmer: This should be covered in f4bf949.

pyfarm/agent/sysinfo/network.py
opalmer: This should be covered in 511941a

opalmer:
@guidow, PTAL

PR #351, opened 2015-11-03
Build documentation as Travis step

opalmer:
Our current Travis setup makes it easy to break the documentation without knowing it.  For example right now it's broken because we're using an old psutil api which no longer exists.  This PR fixes the documentation problem and adds a build step so we don't have this issue in the future.


opalmer:
@guidow, PTAL

guidow:
Looks good to me.

PR #352, opened 2016-01-28
Fix date format in test_uses_provided_create_time

guidow:


opalmer:
Thanks, this looks good to me.  Probably should merge this after we fix the build in #353.

PR #353, opened 2016-01-29
Fix The Build: Drop support for 2.6, get our own copy of run_with_compiler.cmd

opalmer:
From 77ae618:

```
Removing of Python 2.6 from our tests
has been discussed before and it's something
we're already doing on the master side.  The other
reason we should do this is Twisted itself no longer
supports Python 2.6:

  https://twistedmatrix.com/pipermail/twisted-python/2015-November/029993.html
```

From 10fa922:

```
The remote copy has been moved and since it's not
changing there's not really any reason to not keep our
own copy.  This will also be more reliable because we
don't have to make an extra web request.
```


opalmer:
@guidow, PTAL

guidow:
Looks good to me.

PR #354, opened 2016-01-30
Codecov Integration

opalmer:
[codecov.io](https://codecov.io) is a service which provides an easy way to view coverage results.  While we already use [coveralls.io](https://coveralls.io) there's a few things which codecov.io offers that can improve on our experience with code coverage:
- It works on Windows, Mac and Linux.
- It easily integrates with both Travis and AppVeyor.
- Coverage reports are commit dependent rather than build dependent.
- Codecov will make one comment in a review and update it rather than making several.
- The results are shown faster and the interface is cleaner.

### DIFF
--- a/.ci/travis/install.sh
+++ b/.ci/travis/install.sh
@@ -16,6 +16,7 @@ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
 fi
 
 pip install --upgrade pip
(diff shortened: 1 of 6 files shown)


opalmer:
@guidow, PTAL.  I've been using this on another project, [pywincffi](https://github.com/opalmer/pywincffi), for a while and overall it's a much better service.

